PR #3, opened 2020-01-13
Prevent test_cnn.py from crashing w/invalid Tensor label

itsderek23:
When an object is detected and then leaves the image, the Tensor label returned is typically far outside the list of label indexes (ex: 7552.0). This causes the `test_cnn.py` to crash.

This checks to ensure the index is within the available labels before applying the label text.

The exception:

```
  File "test_cnn.py", line 85, in <module>
    cv2.putText(frame, labels[int(e[0]['label'])], pt_t1, cv2.FONT_HERSHEY_SIMPLEX, 0.5, (0, 0, 255), 2)
IndexError: list index out of range
```

### DIFF
--- a/python-api/test_cnn.py
+++ b/python-api/test_cnn.py
@@ -82,10 +82,15 @@ while True:
                     cv2.rectangle(frame, pt1, pt2, (0, 0, 255))
 
                     pt_t1 = x1, y1 + 20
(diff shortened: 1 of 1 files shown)


Luxonis-Brandon:
Thanks man!  I just tried to tag Yurii but his handle isn’t auto-populating on my phone for some reason. Will ping him in slack.

PR #12, opened 2020-02-05
Adding autostart functionality

Luxonis-Brian:
/autostart/ contains a bash script that kicks off startupdemo.py. The rasbian image is configured to automatically run the bash script in /autostart/, so this should allow us to configure the autostart experience via github without reflashing everything.

### DIFF
--- a/autostart/startupdemo
+++ b/autostart/startupdemo
@@ -1,0 +1,16 @@
+#!/bin/bash
+sleep 1
+echo Loading OpenVINO Paths
+source /opt/intel/openvino/bin/setupvars.sh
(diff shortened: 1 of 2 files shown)


Luxonis-Brian:
Note that this will not run unless the /home/pi/.config/autostart/runai.desktop exists and is configured. 

The contents of runai.desktop:

```
[Desktop Entry]
Encoding=UTF-8
Name=RunAI
Exec=lxterminal -e "/home/pi/Desktop/DepthAI/autostart/startupdemo"
Icon=lxterminal
Type=Application
Categories=Utility;
```

PR #14, opened 2020-02-05
Ability to `pip install -e .` so modules can be accessed the repo parent directory

itsderek23:
If a user wants to write their own script that relies on DepthAI, they currently need to create it within the top-level of this repo. This is painful:

* Since this is checked out as a git repo, it's easy to mistakenly add the file to our repo and not your own repo. 
* It's awkward to tell someone that they need to `cd` into their locally checked out copy

This PR adds a workaround so folks can access depthai and its dependencies globally. 

The process on a local computer:

```
cd depthai-python-extras
pip3 install -e .
```

Then from anywhere on the host they can access depthai:

```
$ python3
>>> import depthai
>>> depthai
<module 'depthai' from '/home/pi/projects/depthai-python-extras/depthai.cpython-37m-arm-linux-gnueabihf.so'>
>>> import consts.resource_paths
>>> consts
<module 'consts' from '/home/pi/projects/depthai-python-extras/consts/__init__.py'>
>>> consts.resource_paths.device_cmd_fpath
'/home/pi/projects/depthai-python-extras/depthai.cmd'
>>> import sys
>>> sys.path
['', '/opt/intel/openvino/python/python3.7', '/opt/intel/openvino/python/python3', '/opt/intel/openvino/deployment_tools/model_optimizer', '/home/pi', '/usr/lib/python37.zip', '/usr/lib/python3.7', '/usr/lib/python3.7/lib-dynload', '/home/pi/.local/lib/python3.7/site-packages', '/home/pi/projects/depthai-python-extras', '/usr/local/lib/python3.7/dist-packages', '/usr/lib/python3/dist-packages']
>>> 
```

The `e` flag installs an editable python package that adds this repo's path to `sys.path`. Once it is added to the path, we're able to load modules. 

You can see it via `pip list`:

```
pip list | grep depth
depthai-extras    0.0.0       /home/pi/projects/depthai-python-extras

```

This is an interim workaround to access depthai system-wide until we have a proper `pip`-distributed python package.


itsderek23:
This is a more manual way to do it, but would require adding a line to top of every script or doing some unexpected-magic within our own files.

https://stackoverflow.com/questions/4383571/importing-files-from-different-folder

itsderek23:
Tory verified this worked on his device.

Luxonis-Brandon:
Fantastic!  Thanks, both!

Issue #23, opened 2020-02-15
Add `--config_overwrite` option to test.py


itsderek23:
The `test.py` script doesn't have an argument parsing like [calibrate.py](https://github.com/luxonis/depthai-python-extras/blob/master/calibrate.py#L69), which means adjusting the config must be done in code. This presents a couple of problems:

* During development, it's easy to forget to switch config values back to sensible defaults (see #22).
* The [1098obc](http://localhost:4000/products/bw1098obc/) has a distance of 7.5cm between the stereo cameras, which means currently they'd have to modify the `test.py` code to set the correct distance.
* If you want to verify the stereo calibration, you have to manually modify the code to adjust the streams.

I suggest adding a `--config_overwrite` option to test.py that behaves the same as `calibrate.py` for consistency.

itsderek23:
Closed by #24.

PR #24, opened 2020-02-17
Adds --config-overwrite to test.py

itsderek23:
This adds the ability to overwrite the default config via a script argument. It works the same as the calibrate.py.

Additionally:

* Puts utils in a dedicated `depthai_helpers` folder
* For utils shared by both calibrate and test, adds a depthai_helpers.utils module

### DIFF
--- a/calibrate.py
+++ b/calibrate.py
@@ -1,5 +1,6 @@
 import depthai
-from  calibration_utils import *
+from depthai_helpers.calibration_utils import *
+from depthai_helpers import utils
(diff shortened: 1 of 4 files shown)


itsderek23:
Implements #23

itsderek23:
Example usage for the USB3 w/onboard cameras:

```
python3 test.py -co '{"board_config": {"left_to_right_distance_cm": 7.5}}'
```

PR #27, opened 2020-02-18
Reduce latency for previewout (RGB)

itsderek23:
This reduces the latency for the previewout stream.

Needs testing on RPi Compute by @itsderek23.


itsderek23:
This is working great for me.

I've merged master into this branch as well as it was a bit behind.

itsderek23:
Thanks @alex-luxonis!

PR #37, opened 2020-02-26
Add support for colorized, grayscale, and raw depth map

GergelySzabolcs:
By default, the `COLORMAP_HOT` display is used.

### DIFF
--- a/test.py
+++ b/test.py
@@ -23,8 +23,8 @@ def parse_args():
     ## USB3 w/onboard cameras board config:
     python3 test.py -co '{"board_config": {"left_to_right_distance_cm": 7.5}}'
 
-    ## Show the depth stream:
-    python3 test.py -co '{"streams": ["depth_sipp"]}'
+    ## Show the left+depth stream:
+    python3 test.py -co '{"streams": ["left","depth_sipp"]}'
     '''
     parser = ArgumentParser(epilog=epilog_text,formatter_class=argparse.RawDescriptionHelpFormatter)
     parser.add_argument("-co", "--config_overwrite", default=None,
--- a/test.py
+++ b/test.py
@@ -199,10 +199,13 @@ while True:
                     cv2.putText(frame, "fps: " + str(frame_count_prev[packet.stream_name]), (25, 50), cv2.FONT_HERSHEY_SIMPLEX, 1.0, (0, 0, 255))
                     cv2.imshow(packet.stream_name, frame)
                 else: # uint16
-                    frame_u8 = (frame // 256).astype(np.uint8)
-                    cv2.putText(frame_u8, packet.stream_name, (25, 25), cv2.FONT_HERSHEY_SIMPLEX, 1.0, 255)
-                    cv2.putText(frame_u8, "fps: " + str(frame_count_prev[packet.stream_name]), (25, 50), cv2.FONT_HERSHEY_SIMPLEX, 1.0, 255)
-                    cv2.imshow(packet.stream_name, frame_u8)
+                    frame = (65535 // frame).astype(np.uint8)
+                    #colorize depth map, comment out code below to obtain grayscale
+                    frame = cv2.applyColorMap(frame, cv2.COLORMAP_HOT)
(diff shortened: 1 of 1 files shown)


itsderek23:
@GergelySzabolcs - thanks. Is the color depth map the default?

itsderek23:
Ha ... saw that is in the truncated portion of commit message. thx!

itsderek23:
@GergelySzabolcs - I'm seeing 3-5 second latency w/COLORMAP_JET+COLORMAP_HOT on the RPi Compute. Grayscale has much lower latency.

@Luxonis-Brandon @Luxonis-Brian - what's an acceptable latency, in your opinion? The colorized options feels pretty slow as a default.

👉  __[🎥 Video sample](https://drive.google.com/file/d/1wqS9I12TrtzVu_YbR0S-5MmIxFw4aHS-/view?usp=sharing).__

GergelySzabolcs:
That's what the PI is capable of in terms of applying a color LUT. @itsderek23

Luxonis-Brandon:
I think the previous implementation was much faster.  I did take some video of it back then and will share below.

So here's what the depth_color_h looked like as of February 3rd, 3:51pm Colorado time (GMT-7:00, in case that helps for tracking down the commits):

[https://photos.app.goo.gl/EdgWJMAasHudjjRe9](https://photos.app.goo.gl/EdgWJMAasHudjjRe9)
[https://photos.app.goo.gl/ANqghT4KNXt9bEXA6](https://photos.app.goo.gl/ANqghT4KNXt9bEXA6)
[https://photos.app.goo.gl/cdMQmr3bdEuAT4VVA](https://photos.app.goo.gl/cdMQmr3bdEuAT4VVA)

Seemed to be about 7-8FPS, but I'm unsure on the actual latency, but about 1 second maybe?

So I'm thinking the way the format was passed from the Myriad X makes it easier/faster for the Pi to convert it?

This was run on a BW1097 - so the Pi Compute Module 3+, which is the same processor/specs as the Raspberry Pi 3B+ Pi.

This is the same hardware that @itsderek23 is testing on as well.

So I think we need to do what's necessary to get back to this performance we saw with this Feb. 3rd version of the codebase.  This new version of colorized depth (video [here](https://drive.google.com/file/d/1wqS9I12TrtzVu_YbR0S-5MmIxFw4aHS-/view?usp=sharing)) is too slow to be usable for checking calibration and how the depth picking up objects.  

The 7-8FPS w/ whatever latency it is seems totally fine for this purpose however.

Thoughts?

Thanks,
Brandon

GergelySzabolcs:
Agreed. I'm on my way of debugging.

Luxonis-Brandon:
Perfect, thanks @GergelySzabolcs !

itsderek23:
@GergelySzabolcs found that performance was better when running bc91e1f60b1fb3549994590651621fdab763acb3 w/our current internal depthai `devices_auto_reset` branch. This PR is actually identical to that commit.

I'm able to get a low-latency `depth_sipp` stream by also running a `left` stream:

```
python3 test.py -co '{"streams": ["left","depth_sipp"]}'
```

The high-latency when using just the `depth_sipp` stream might be caused by the high frame rate triggered by using a single stream. I'll see FPS > 30 w/the single stream vs. 5 FPS when using both streams. For testing calibration results, low latency > high FPS. My hunch is the host-side Python logic for adding color adds significant overhead because it is being applied to a higher number of frames. I'll also see "data queue is full" which happens when the Python code can't keep up.

Next steps:

* Update `--help` output to include the `left` stream in the sample for calibration
* Update our calibration docs to use the left stream as well. 

We have an internal `FPS limit` branch in progress that might help us with displaying just the single stream w/low latency, but using 2 streams seems like an acceptable workaround.

Does this make sense @Luxonis-Brian and @Luxonis-Brandon?

itsderek23:
To use `JET` color mapping:

* [Comment out L204 of test.py](https://github.com/luxonis/depthai-python-extras/blob/master/test.py#L204)
* Uncomment L205

...if the usage of different depth visual outputs becomes common (or code toggling becomes painful), we can make this an argument when running the script.

PR #38, opened 2020-02-28
Colorized depth map + fps limit and multi tensor output 

GergelySzabolcs:
This pr contains:
-multi tensor output support
-fix for occasional crashes


itsderek23:
Thanks @GergelySzabolcs!

>  -fix for occasional crashes

Clarifying: does this addresses crashes you've seen on our current `master` branch, or crashes introduced in development of this branch?

Luxonis-Brandon:
Yes, thanks @GergelySzabolcs !  

>  rebuild libs; limit previewout, depth sipp to 10 fps

So for our previewout-only example script (the default test.py), will this now be limited to 10FPS by default as well?  

I ask because it's compelling initial demo for that to show the full 30FPS.  Although 10FPS isn't bad, so that might be fine as well.  

Either way, just wanting to make sure I understand.

Thanks again!

GergelySzabolcs:
> Thanks @GergelySzabolcs!
> 
> > -fix for occasional crashes
> 
> Clarifying: does this addresses crashes you've seen on our current `master` branch, or crashes introduced in development of this branch?

Crashes seen on master

GergelySzabolcs:
> Yes, thanks @GergelySzabolcs !
> 
> > rebuild libs; limit previewout, depth sipp to 10 fps
> 
> So for our previewout-only example script (the default test.py), will this now be limited to 10FPS by default as well?
> 
> I ask because it's compelling initial demo for that to show the full 30FPS. Although 10FPS isn't bad, so that might be fine as well.
> 
> Either way, just wanting to make sure I understand.
> 
> Thanks again!

Limiting the fps it is supposedly just an example of usage. 
We should decide which streams we enable by default. 
In test.py depth aipp, previewout and metaout are enabled, that's why the limiting to 10 fps, so we don't saturate usb2 bandwidth. If depth sipp is not enabled we can do 30 fps.

Luxonis-Brandon:
Fantastic.  Thanks.  

This is a great feature to have - the capability to limit the framerate so as to not saturate USB2 bandwith.

I'm going to move this to complete on the roadmap.  Thanks for knocking out this feature in the process of solving the calibration/colorized-depth display problem.

For the default display:
I'm thinking we stick w/ previewout w/ the metadata visualized on it.  As it's nice and 30FPS even on a Pi w/ USB2.  CC: @itsderek23 for any thoughts on this.

Thanks again!

test.py
```diff
@@ -1,81 +1,53 @@
+import json
 import sys
 from time import time
 from time import sleep
-import argparse
```
itsderek23: It looks like the `test.py` you were working from was pretty dated. It doesn't contain argument parsing, so users cannot overrwrite the config file via an argument.

test.py
```diff
@@ -1,81 +1,53 @@
...
 import numpy as np
 import cv2
+import struct
 
 import depthai
 
 import consts.resource_paths
-from depthai_helpers import utils
-
-def parse_args():
```
itsderek23: See earlier comment about removing argument parsing (this should remain in the script)

test.py
```diff
@@ -1,81 +1,53 @@
...
-    # ['left', 'right','previewout', 'metaout', 'depth_sipp']
-    # If "left" is used, it must be in the first position.
-    # To test depth use:
-    # ['metaout', 'previewout', 'depth_sipp']
-    'streams': ['metaout', 'previewout'],
-    'depth':
+
+# Make sure to put 'left' always first. Workaround for an issue to be investigated
+configs = {
+    'streams': [
```
itsderek23: Any reason we're showing all of the streams by default? To keep things simple, we just used ['metaout', 'previewout' previously.

test.py
```diff
@@ -89,19 +61,23 @@ def parse_args():
     },
     'board_config':
     {
-        'swap_left_and_right_cameras': True, # True for 1097 (RPi Compute) and 1098OBC (USB w/onboard cameras)
-        'left_fov_deg': 69.0, # Same on 1097 and 1098OBC
-        'left_to_right_distance_cm': 9.0, # Distance between stereo cameras
-        'left_to_rgb_distance_cm': 2.0 # Currently unused
+        'swap_left_and_right_cameras': False,
```
itsderek23: Any reason comments were dropped and the default values were changed? We want the defaults to line up with the BW1097, which is the most popular edition of DepthAI.

test.py
```diff
@@ -89,19 +61,23 @@ def parse_args():
...
+        'left_fov_deg': 69.0,
+        'left_to_right_distance_cm': 3.5,
+        'left_to_rgb_distance_cm': 0.0
     }
 }
 
-if args['config_overwrite'] is not None:
```
itsderek23: This should remain (artifact of working on an old test.py, I think).

test.py
```diff
@@ -89,19 +61,23 @@ def parse_args():
...
-    config = utils.merge(args['config_overwrite'],config)
-    print("Merged Pipeline config with overwrite",config)
+if 'depth_sipp' in configs['streams'] and ('depth_color_h' in configs['streams'] or 'depth_mm_h' in configs['streams']):
```
itsderek23: I like that check is included

test.py
```diff
@@ -158,6 +134,9 @@ def parse_args():
                 if e[0]['confidence'] > 0.5:
                     x1 = int(e[0]['left'] * img_w)
                     y1 = int(e[0]['top'] * img_h)
+                    # print(e[0]['distance_x'])
```
itsderek23: Should we just remove these commented-out lines?

test.py
```diff
@@ -173,15 +152,15 @@ def parse_args():
 
                         pt_t2 = x1, y1 + 40
                         cv2.putText(frame, '{:.2f}'.format(100*e[0]['confidence']) + ' %', pt_t2, cv2.FONT_HERSHEY_SIMPLEX, 0.5, (0, 0, 255))
+                        if configs['ai']['calc_dist_to_bb']:
```
itsderek23: Mind providing a description for what `calc_dist_to_bb` means?

example_run_age_gender_recognition_retail.py
```diff
@@ -0,0 +1,115 @@
+import sys
```
itsderek23: This example runs for me. 👍

example_run_landmarks_recognition.py
```diff
@@ -29,7 +29,8 @@
 
         # 35 landmarks
         'blob_file_config': consts.resource_paths.prefix + 'nn/object_recognition_4shave/landmarks/landmarks-config-35.json',
-        'blob_file': consts.resource_paths.prefix + 'nn/object_recognition_4shave/landmarks/facial-landmarks-35-adas-0002.blob'
+        'blob_file': consts.resource_paths.prefix + 'nn/object_recognition_4shave/landmarks/facial-landmarks-35-adas-0002.blob',
```
itsderek23: This works for me. 👍

example_run_vehicle_attributes_recognition_barrier.py
```diff
@@ -0,0 +1,142 @@
+import sys
```
itsderek23: This runs for a bit, but classifies me as a red car when I'm wearing a blue sweatshirt.

After a minute or so, it crashes with:

```
ValueError: array is too big; `arr.size * arr.dtype.itemsize` is larger than the maximum possible size.
  File "example_run_vehicle_attributes_recognition_barrier.py", line 121, in <module>
    data0 = data[0,:,:]
IndexError: too many indices for array
Stopping threads: ...
```

resources/nn/object_recognition_4shave/age_gender_recognition_retail/age-gender-recognition-retail-0013.json
```diff
@@ -0,0 +1,65 @@
...
+                ["female", "male"]
+            ],
+            "property_value_mapping":
+            [
+                []
+            ],
+            "output_properties_type": "f16"
+       }
+    ],
+    "mappings":
```
itsderek23: Why are these labels included?

resources/nn/object_recognition_4shave/vehicle_attributes_recognition_barrier/vehicle-attributes-recognition-barrier-0039.json
```diff
@@ -0,0 +1,65 @@
...
+            "property_value_mapping":
+            [
+                []
+            ],
+            "output_properties_type": "f16"
+       }
+    ],
+    "mappings":
+    {
+        "labels":
```
itsderek23: Why are these labels included?

test.py
GergelySzabolcs: calc_dist_to_bb is enabling or disabling depth calculation for bounding box (some models don't have bounding box, or some HW don't have stereo cams

example_run_vehicle_attributes_recognition_barrier.py
GergelySzabolcs: This one was done by yurii.
I just merged his latest work with `multiple_outpout_tensors` branch.

resources/nn/object_recognition_4shave/vehicle_attributes_recognition_barrier/vehicle-attributes-recognition-barrier-0039.json
GergelySzabolcs: I don't know. It was done by yurii.

GergelySzabolcs:
@itsderek23  i just updated the branch to match master. Also merged yurii's multiple output tensors branch, which might contain fixes for models that are not working?

test.py
itsderek23: Thanks!

example_run_vehicle_attributes_recognition_barrier.py
itsderek23: Inference still seems incorrect:

image

...I am pale/white for sure, but not a car :).

Good news is it hasn't crashed for me over a period of a couple of minutes ... going to let it run longer.

itsderek23:
@GergelySzabolcs - test.py still freezes for me on the 1097 after showing the first frame w/data queue full for metaout/preview. I haven't yet dug through the host-side Python code to see if we've introduced any performance changes there.

example_run_vehicle_attributes_recognition_barrier.py
Luxonis-Brandon: I’d be curious if we fed that same image into the same network running on a host computer for inference it it would have the same output.

itsderek23:
Yes - I wouldn't be shocked if this was an issue w/the NN itself.

This does appear to be stable now though. Thanks!

Luxonis-Brandon:
Yes.  Could be.  

I think we completely punt on it for now and remove it from any all current efforts and planned future efforts and/or tests.

With the focus on the following instead:

1. Object detector models that don't have any additional/special metadata.  So that normally just return class and bounding box locations.
2.  Augmentation of these object detector models with disparity depth results - to allow 3D object localization (i.e. the object class, and the XYZ locations of the centroid of the bounding box).
3.  Check of calibration - so the capability to visualize the depth map on a Raspberry Pi 3B+ with JET color map.  The check of calibration is necessary for users to enable 2.

So I'm thinking we focus all PRs/efforts going forward to cover those until we get that dialed.  

Once we have that dialed, we can move on to support of other network types, multi-output-tensors, etc.  But I'm thinking we should really make it a conscious decision, with a check-off list that those are dialed.

Thoughts?

itsderek23:
Superseded by #41.

PR #42, opened 2020-03-06
Load depthai-api repo as a submodule

itsderek23:
This loads the `depthai-api` Git repo as a submodule.

The `depthai-api` repo contains the host-side code that is used to build the Python `.so` file. End users do not need to grab this submodule unless they are building the `so` file from source.

Instructions for building the `so` file will be within the `depthai-api` repo README.

Run the following to grab the submodule:

```
git submodule update --init
```

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "depthai-api"]
 	path = depthai-api
-	url = ../depthai-api
+	url = https://github.com/luxonis/depthai-api.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "depthai-api"]
+	path = depthai-api
+	url = ../depthai-api
--- a/README.md
+++ b/README.md
@@ -6,28 +6,43 @@ __Documentation is available at [https://docs.luxonis.com](https://docs.luxonis.
 
 ## Python modules
 
-Files with extention `.so` are python modules:  
+Files with `.so` extension are the python modules:  
 - `depthai.cpython-36m-x86_64-linux-gnu.so` built for Ubuntu 18.04 & Python 3.6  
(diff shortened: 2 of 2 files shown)


Luxonis-Brandon:
Super exciting that this is progressing!  Will be great for users to be able to build the host side on their own custom platforms/OS/dependencies/etc.

alex-luxonis:
Referenced the [depthai-api](https://github.com/luxonis/depthai-api) submodule in README.md and added basic instructions to init and build.
Not using `--recursive` for submodule init, as that would fetch some large (and unused by us) modules in opencv/dldt.

itsderek23:
I tested w/a fresh checkout on the RPi Compute. It builds w/o any options for the `make` jobs.

Two issues I see after building and running `git status`:

```
modified:   depthai-api (untracked content)
modified:   depthai.cpython-37m-arm-linux-gnueabihf.so
```

## modified:   depthai-api (untracked content)

To get rid of this, I wonder if we need to do the following [via Stackoverflow](https://stackoverflow.com/a/52722775):

```
git rm -rf --cached depthai-api
git add depthai-api
```

## modified:   depthai.cpython-37m-arm-linux-gnueabihf.so

I wonder if the binary file is different because of merging #43? I'm concerned about losing the `.so` files updates in #43 after merging. Is the safest thing running `./depthai-api/build_py_module.sh` after merging to generate a new `so`, then committing that?

alex-luxonis:
```
modified:   depthai.cpython-36m-x86_64-linux-gnu.so
```
We didn't update the x86_64 lib for the latest `depthai-api` changes related to USB3. Maybe we should do this for consistency? (for now only the RPi lib is updated).
But anyway, the file may appear `modified` after a rebuild, even if built with the same source code, as some absolute paths from the build machine get embedded into the binary.


As for the untracked content in `depthai-api`, could you check:
```
depthai-api$ git status
```
I don't get this on my machine (the `build` dir should be already in .gitignore

PR #53, opened 2020-04-08
Check if udev rules are set

szabi-luxonis:
Check if udev rules are set and fix quit on 'q' keypress.


Luxonis-Brandon:
So checking if the USB rules are set works great!  The flashing message is nice too.  

image


One improvement is probably worthwhile though: stop attempting to connect (and thereby stop showing the message) after it's determined once that the USB rules are not set.

See the video below:
[https://photos.app.goo.gl/Cfb6pYKu4vmzVxT38](https://photos.app.goo.gl/Cfb6pYKu4vmzVxT38)

Background:
So the first messages stopped where you could read it, but then after say a second the screen started spamming more messages, making it hard to read.

Stopping after the first message would make this awesome.  I would assume not hard as well?

Thoughts?

Thanks,
Brandon

szabi-luxonis:
fixed, @Luxonis-Brandon

Luxonis-Brandon:
Thanks, just tested and this LGTM.  I think it's good to merge @szabi-luxonis !

image

Luxonis-Brandon:
Oh one more thing @szabi-luxonis .  So after running those commands above, it's necessary to unplug DepthAI from USB and plug it back in.  So we should change the error message to say:

> To correct this, please:
> 1.  Run the following commands to set the udev rules for DepthAI:
> echo 'SUBSYSTEM=="usb", ATTRS{idVendor}=="03e7", MODE="0666"' | sudo tee /etc/udev/rules.d/80-movidius.rules
> sudo udevadm control --reload-rules && udevadm trigger
> 2.  Unplug device USB.
> 3.  Plug device USB back in.

If you don't unplug/replug you see these errors:
image

And after unplugging and replugging it successfully runs:
image

Luxonis-Brandon:
This LGTM thanks @szabi-luxonis !

PR #56, opened 2020-04-09
Save calibration & board config to EEPROM

alex-luxonis:
New options added to `depthai.py`:
```
  -fv FIELD_OF_VIEW, --field-of-view FIELD_OF_VIEW
                        Field of view for the stereo cameras in [deg]
  -b BASELINE, --baseline BASELINE
                        Left/Right camera baseline in [cm]
  -r RGB_BASELINE, --rgb-baseline RGB_BASELINE
                        Distance the RGB camera is from the Left camera
  -w, --no-swap-lr      Do not swap the Left and Right cameras
  -e, --store-eeprom    Store the calibration and board_config (fov,
                        baselines, swap-lr) in the EEPROM onboard
  --clear-eeprom        Invalidate the calib and board_config from EEPROM
  -o, --override-calib  Use the calib and board_config from host, ignoring the
                        EEPROM data if programmed
```

New options added to `calibrate.py`:
```
  -fv FIELD_OF_VIEW, --field-of-view FIELD_OF_VIEW
                        Field of view for the stereo cameras in [deg]
  -b BASELINE, --baseline BASELINE
                        Left/Right camera baseline in [cm]
  -w, --no-swap-lr      Do not swap the Left and Right cameras
```

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2019 Luxonis-Brandon
+Copyright (c) 2020 Luxonis-Brandon
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
--- a/calibrate.py
+++ b/calibrate.py
@@ -70,8 +70,8 @@ def parse_args():
     parser.add_argument("-co", "--config_overwrite", default=None,
                         type=str, required=False,
                         help="JSON-formatted pipeline config object. This will be override defaults used in this script.")
-    parser.add_argument("-fv", "--field-of-view", default=69.0, type=float,
-                        help="Field of view for the stereo cameras in [deg]")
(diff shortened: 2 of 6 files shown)


Luxonis-Brandon:
Thanks @alex-luxonis !  So quick question on field of view... is that diagonal or horizontal?

Luxonis-Brandon:
So other than specifying which FOV this means, this LGTM.  And @Luxonis-Brian checked it out and it LGTH as well.  So I'd say feel free to go with this @alex-luxonis !

Luxonis-Brian:
The horizontal FOV should actually be 71.86, and the vertical should be 56.74.

alex-luxonis:
Note `depthai_usb2.cmd` was not rebuilt here!

Luxonis-Brandon:
Should we do that for now @alex-luxonis or should we do one of the approaches for supporting USB2-only that you or Szabi mentioned the other day?

PR #59, opened 2020-04-13
Stereo sync L/R, apply same exposure

alex-luxonis:
- Synchronize Left and Right cameras by using broadcast start
- Apply Left camera exposure to both Left and Right


Luxonis-Brandon:
LGTM.

- Tested stereo and force USB2 `-fusb2` mode on BW1097.  Did the force USB2 mode intentionally on BW1097 (which is only USB2) to make sure that such a force on a USB-only host didn't do something unexpected.  It worked fine.
- Tested `-fusb2` on USB3 Ubuntu 18.04 host on USB3 port and it worked properly as well.  I like the flashing warning.  (USB3 also still worked fine when `-fusb2` was not specified.

Luxonis-Brian:
LGTM.

I calibrated a fresh 1097 and tested `depth_sipp` to make sure I didn't see the up/down motion artifacts. I did not see them, so I believe the broadcast start sync is working as expected. 

I also tested the `-fusb2` flag on the OBC (luxonis-test PC as host, USB3 port) using `depth_sipp` output as well. I saw a drop from 30FPS to ~18 FPS. I then plugged my USB cable into a USB2 port and ran without the `-fusb2`  and found 18FPS as well.

PR #60, opened 2020-04-14
Refactor the calibration script to make it faster and clearer

VanDavv:
This PR improves the overall performance of the calibration script and it's readability and adds some additional functionalities:
- Capture the frames only after spacebar is hit
- Stops the capture if it could not find the chessboard
- Marks the captured polygon green if the image was captured successfully (taken from #51, thanks @squeakus)
- Do not save the images if the chessboard was not found (also taken from #51, thanks again @squeakus)
- removed `--polygons` flag as it was not giving any relevant functionalities to the user.

This is a significant refactor, would love to hear your thoughts and if it works on your side @Luxonis-Brandon  and @Luxonis-Brian

### DIFF
--- a/calibrate.py
+++ b/calibrate.py
@@ -90,8 +90,11 @@ def find_chessboard(frame):
 
 
 class Main:
+    output_scale_factor = 0.5
     cmd_file = consts.resource_paths.device_cmd_fpath
     polygons = None
+    width = None
+    height = None
     current_polygon = 0
     images_captured_polygon = 0
     images_captured = 0
@@ -152,6 +155,47 @@ class Main:
         print("py: Saved image as: " + str(filename))
         return True
 
+    def show_info_frame(self):
+        info_frame = np.zeros((600, 1000, 3), np.uint8)
+        print("Starting image capture. Press the [ESC] key to abort.")
+        print("Will take {} total images, {} per each polygon.".format(self.total_images, self.args['count']))
+
+        def show(position, text):
+            cv2.putText(info_frame, text, position, cv2.FONT_HERSHEY_SIMPLEX, 1.0, (0, 255, 0))
+
+        show((25, 100), "Information about image capture:")
+        show((25, 160), "Press the [ESC] key to abort.")
+        show((25, 220), "Press the [spacebar] key to capture the image.")
+        show((25, 300), "Polygon on the image represents the desired chessboard")
+        show((25, 340), "position, that will provide best calibration score.")
+        show((25, 400), "Will take {} total images, {} per each polygon.".format(self.total_images, self.args['count']))
+        show((25, 550), "To continue, press [spacebar]...")
+
+        cv2.imshow("info", info_frame)
+        while True:
+            key = cv2.waitKey(1)
+            if key == ord(" "):
+                cv2.destroyAllWindows()
+                return
+            elif key == 27 or key == ord("q"):  # 27 - ESC
+                cv2.destroyAllWindows()
+                raise SystemExit(0)
+
+    def show_failed_capture_frame(self):
+        width, height = int(self.width * self.output_scale_factor), int(self.height * self.output_scale_factor)
+        info_frame = np.zeros((height, width, 3), np.uint8)
+        print("py: Capture failed, unable to find chessboard! Fix position and press spacebar again")
+
+        def show(position, text):
+            cv2.putText(info_frame, text, position, cv2.FONT_HERSHEY_TRIPLEX, 0.7, (0, 255, 0))
+
+        show((50, int(height / 2 - 40)), "Capture failed, unable to find chessboard!")
+        show((60, int(height / 2 + 40)), "Fix position and press spacebar again")
+
+        cv2.imshow("left", info_frame)
+        cv2.imshow("right", info_frame)
+        cv2.waitKey(2000)
+
     def capture_images(self):
         finished = False
         capturing = False
@@ -167,11 +211,8 @@ class Main:
                     frame = cv2.cvtColor(frame, cv2.COLOR_GRAY2RGB)
 
                     if self.polygons is None:
-                        height, width, _ = frame.shape
-                        self.polygons = setPolygonCoordinates(height, width)
-                        print("Starting image capture. Press the [ESC] key to abort.")
-                        print("Will take {} total images, {} per each polygon.".format(self.total_images,
-                                                                                       self.args['count']))
+                        self.height, self.width, _ = frame.shape
+                        self.polygons = setPolygonCoordinates(self.height, self.width)
 
                     key = cv2.waitKey(1)
                     if key == ord("q"):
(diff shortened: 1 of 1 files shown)


squeakus:
Hey, this is looking great, the green and red boxes make it clear at the point of capture which one is causing the problem and that is the most important part. I seem to be getting a problem though when it runs the calibration part after image 12:
py: Saved image as: left_p12_12.png

  File "calibrate.py", line 257, in <module>
    Main().run()
  File "calibrate.py", line 250, in run
    self.capture_images()
  File "calibrate.py", line 224, in capture_images
    frame, np.array([self.polygons[self.current_polygon]]),
IndexError: list index out of range

Any idea what might be causing this?

VanDavv:
> Hey, this is looking great, the green and red boxes make it clear at the point of capture which one is causing the problem and that is the most important part. I seem to be getting a problem though when it runs the calibration part after image 12:
> py: Saved image as: left_p12_12.png
> 
> Traceback (most recent call last):
> File "calibrate.py", line 257, in
> Main().run()
> File "calibrate.py", line 250, in run
> self.capture_images()
> File "calibrate.py", line 224, in capture_images
> frame, np.array([self.polygons[self.current_polygon]]),
> IndexError: list index out of range
> 
> Any idea what might be causing this?

I've uploaded a fix, should be fine now

squeakus:
Thanks! I tried it there and that fixed the bug.

The is one more thing though, there is an edge case where it will find the
chessboards in the small frames but will fail to detect them in the larger
ones, the final check before saving the frame will have to be run on the
full size frame otherwise you end up with the program saying it matched
them but when it comes to the final calibration comparison it will throw an
error.  My apologies to keep annoying you!

VanDavv:
@squeakus that's not annoying at all, thanks for the feedback and feel free to do this even more often!

Is it even possible for the algorithm to detect chessboard corners on smaller frame but don't on the larger? Detecting before save will further slow down the algorithm, and I assumed that the mentioned situation is not possible

squeakus:
Yeah, it is really odd and stumped me for ages, why can I save the images
but then it doesn't detect the corners using the exact same algorithm?
Reducing the image size seems to reduce the noise and improve the rate of
detection. I have also found that holding the board further away while
calibrating also has a similar effect. It is quite counter-intuitive.

VanDavv:
Ok, I've just removed the resize at all - there is no sense IMO to first check on small frame and then on larger - especially that I check for this right before save

Luxonis-Brandon:
Wow, thanks both!  I'll give this a try today on my BW1097!

Luxonis-Brian:
Testing on 1097 w/ Raspberry Pi as host - 

I very much like the resized image windows. Makes it much easier to see where the checkerboard is on capturing. 

A few things - 

1. A slight delay that allows the second green polygon to show would be nice. Currently, with a good pair of captures, only one polygon goes green before the polygons switch to the next pose. It would be nice to clearly see both polygons turn green before the next pose is displayed. 

2) At the beginning of the FOR loop, we set `frame=packet.getData()`, but it seems like we don't reset our location within that buffer to the most recent frame. The result is that as we process for the corners, we fall "back in time" in the data queue on the Raspberry Pi, and there's a serious lag once the processor finishes it's findChessboardCorners() call. If we can simply grab the most recent frame from data_list, I think this problem would be solved.

3) The print statement in the case of a failed capture ("stopping capture, unable to find chessboard...") should probably be displayed within the frame as frame text, possibly in addition to in the terminal. 

4) The frame text is super hard to see. Are there any other cv2 fonts or available colors or options to improve visibility? Bolder font maybe? 

5) It seems possible to get the program into a state where it endlessly tries to find corners. I can do this by putting my fingers over the L/R cameras so that it's only gray noise that shows up in each frame. If I press spacebar, the terminal fills up with "Data queue is full" messages, the right and left frame windows freeze up, and I'm unable to 'q' to quit or CTRL+C in the terminal.

VanDavv:
@Luxonis-Brian 

1. we can freeze the window for a second - but I'm not a fan of such solutions, as they force you to see the "delay", whereas you can just move onward quicker without them
2. could you share an example? I could also use this in other projects. The same code was in previous implementation, so didn't change it - which doesn't mean I shouldn't change it.
3. Same as point 1
4. There are for sure, can dig into it
5. this should unfreeze after a while, but I'll improve this 
6. I thought about it for a while and now it's implemented that if you miss one photo, you can position the camera again and take just the one missing. But it's not so good I think in terms of calibration. I'll change it so that when hitting spacebar, it will capture both left and right, and override any previous image created for this polygon

Luxonis-Brian:
1. As it stands now, there is confusing feedback to the user. One polygon clearly turns green, and one stays red before the next polygons are displayed. The terminal tells me that both the left and right images were saved, but were they good? What did the green one mean? Why didn't the other one turn green? Was it bad? In this case, the user feedback is more important than a delay of, for example, 500ms to clearly display both polygons as green. 
2. I uploaded a video in slack regarding this - apologies for not sharing that video here. I didn't see an easy way to upload to this comment section. I definitely could have missed something though...
3. This doesn't have to add any time at all. Just a change to the cv2.putText() displayed on the screen under this case. Keep in mind that this case handling may be tied in with a solution for #6 here, in which you guarantee both captures are good before saving them. 
4. Awesome, thanks! 
5. Maybe it's just the raspberry pi being slow and me being impatient, but I observed this unfreeze once, and it was after several minutes. What's the reason for doing 4 frame captures here, `max_frames_per_capure = 4`? 
6. A stereo calibration requires that both images be taken at the same time so that the 3D objects in the images being used for calibration are the same. This is why both Left and Right have to be captured and verified together and saved together for each polygon. And, actually, now that I think about it, we would want to make sure we're grabbing L/R packets out of the data stream from the same snapshot as well. For example, say the Left packet comes first and then the Right packet. Since those are taken at basically the same time, they'll be microseconds apart in time. However, if we examine the Right packet first and then examine the next Left packet, those might have been captured at 1/FPS seconds apart.

VanDavv:
@Luxonis-Brian I think I addressed all of the points you mentioned.

The font is bolder, I stop iterating over data_list to quicker catch up with the latest frames, I introduced a black screens with info for the user (more convinient than putting them on the received frames IMHO) - check out if you like it, right now only polygon number and images count are shown on the live stream screen

Luxonis-Brandon:
Just trying this out now.  Really like the instructions on the black screen to start.

You know what I just realized would make this so much more intuitive?  If we mirrored to video feeds (for display only).  Apple does this on FaceTime when you show yourself and it makes it so much more intuitive to interact with.  

I think it will make the cal easier too.

Thoughts?

Luxonis-Brandon:
Also on the Pi we should set the framerate to 10 FPS so that the latency stays low.

Luxonis-Brian:
> Also on the Pi we should set the framerate to 10 FPS so that the latency stays low.

Is there an easy way to do this in the code? To recognize the RPI host to change frame rate? 

Otherwise, in the interest of keeping the PRs small and specific, @VanDavv suggested we go ahead and merge this PR, and we can update with @themarpe's synced code later. I'll give this the official LGTM. 

@Luxonis-Brandon?

Luxonis-Brandon:
Yes from here it’s just:
import os
os.uname()

https://raspberrypi.stackexchange.com/questions/5100/detect-that-a-python-program-is-running-on-the-pi

Luxonis-Brandon:
So pulling nodename and matching 'raspberrypi' is all that would be needed, per below.

IMG_7895

And here's an example of using the streams with limited framerate:
https://github.com/luxonis/depthai/blob/f8f6a7f261e548a8a1fbf27dccac3dc835d883cb/depthai.py#L126

Thoughts?

Luxonis-Brandon:
But anyways, let's do that as a separate PR and merge this.  Agreed @Luxonis-Brian .  LGTM.

Luxonis-Brandon:
Documented the two additional features that came out of this below:
https://github.com/luxonis/depthai/issues/64
https://github.com/luxonis/depthai/issues/63

PR #61, opened 2020-04-14
DepthAI support for macOS

alex-luxonis:


Luxonis-Brandon:
So this LGTM to me @alex-luxonis .  Once we integrate the variable time-out after boot I think this is good to merge.

Thoughts?

Luxonis-Brandon:
So @alex-luxonis  - FWIW on my iMac the macos_build branch works fine.  Not using the macos_build_wdog_6s branch.  So I think we push macos_build to Master and then do the reset improvement as a separate effort.

As it seems its an independent issue, and the purpose of macos_build is to get DepthAI building correctly on Mac, and it definitely does that.

In short: LGTM, I think we should push this to Master.

PR #65, opened 2020-04-16
autostart fix for repo name change

Luxonis-Brian:
Simple change here, accompanied by a change in /home/pi/.config/autostart/runai.desktop in Raspbian, which will need to be updated manually for users who have an existing uSD with the old depthai-python-extras repo. All future images on uSDs that ship from Luxonis will have this change.


Luxonis-Brandon:
LGTM.

PR #66, opened 2020-04-17
Frame metadata + calibration improvement

Luxonis-Brian:
@themarpe, your changes worked well. I was able to check that two adjacent packets were less than 1ms difference in time stamps. That seemed to be good enough to prevent L/R captures from different frames being used for a calibration pair. I also added a python list "buffer" that acts to pick out the two most recent data packets, rather than working up from the buffered data stream. The data stream doesn't seem to be subscriptable, so I couldn't find an easy way to select the two most recent items. 

I'm sure there's something obvious I didn't try :)

I pushed my stuff up on top of yours since it was easy, but if there's a better way to do this, please let me know and I'll do that. Apologies for not knowing the conventions here.


VanDavv:
@Luxonis-Brian I left my review on #68 , would be great to have your feedback on it - initialy wanted to do the comments and explain how I would do it, but the same amount of time it took me to actually write it

VanDavv:
Since #68 was merged, this now is LGTM

Luxonis-Brian:
LGTM too. Think this can be merged into master.

Luxonis-Brian:
@themarpe, in testing your latest changes the 1ms timestamp difference we had been using was no longer large enough to reliably capture a frame pair when pressing spacebar. I didn't test to find the threshold but a 10ms difference did work. 

This is on the 1097 compute module.

VanDavv:
@Luxonis-Brian should me merge this?

alex-luxonis:
There is an issue with Left/Right sync as @Luxonis-Brian pointed above. We'll work to resolve this first.

themarpe:
Rebuilt the binaries. If everyone is okay with the changes I'll merge with master.

Luxonis-Brandon:
Testing this now.

Luxonis-Brandon:
So in testing this it seems the -iv and -ih flags do not seem to be honored from https://github.com/luxonis/depthai/pull/71 . I'm guessing this is the conflict with calibrate.py?  Looking now.

Luxonis-Brandon:
OK this looks great to me!  It works properly on the BW1097.  In testing this I discovered calibrate.py crashes on Mac OS X (https://github.com/luxonis/depthai/issues/78) for this version and also Master - so not an issue with this PR.

LGTM.

Luxonis-Brian:
Just checked that the 1ms time stamp flag was working, and it is. I think we're good here. Thanks @themarpe.

szabi-luxonis:
@themarpe please open PR on depthai-api too

PR #68, opened 2020-04-17
propose another aproach to timestamp measurements

VanDavv:
I noticed some things I didn't like on PR #66 

And instead of proposing a new solution, I just wrote one quick, easier for me to operate on a real code instead of suggestions

Key issues I found or changes I made

- first, you check metadata on `packet_list[-2]` and `packet_list[-1]` (last frames in array), whereas to `packet_pair` you take `packet_list[0]` and `packet_list[1]`

- There is an assumption that the last two items in `packet_list` are most recent ones

- There may be a case that the `packet_pair` will consist of two left camera frames (we just assume they won't come so fast to pass the timestamp check)

Also, creating two arrays (`packet_list` and `packet_pair`), just to get two most recent left and right frames, was something I'd like to avoid - I always tend to stick to variables wherever I can instead of arrays, both from readability and performance reasons.

Main change here is that I introduced `recent_left` and `recent_right` variables that store the most recent frames from each stream. When a new `data_list` arrives, I iterate over it and compare whether it has more recent timestamp than previous one, and if it does, the packet becomes either `recent_left` or `recent_right`.

Then, before capturing, I just check their timestamp diff to verify whether they're good to go or not.

One further improvement I can see here is that we can find a pair, in the `data_list` stream, that will have the timestamp diff acceptable, but they won't be the recent ones - and it doesn't really matter if they are, as long as their time diff is in the threshold. 
So instead of checking the `data_list` for the most recent ones, we can try to find the minimum timestamp diff pair.

Thoughts @Luxonis-Brian ?

EDIT
I've added #69 where I implemented this further improvement - and still I need to test both this PR and the #69

### DIFF
--- a/calibrate.py
+++ b/calibrate.py
@@ -86,6 +86,10 @@ def find_chessboard(frame):
            cv2.findChessboardCorners(frame, (9, 6), chessboard_flags)[0]
 
 
+def ts(packet):
+    return packet.getMetadata().getTimestamp()
+
+
 class Main:
     output_scale_factor = 0.5
     cmd_file = consts.resource_paths.device_cmd_fpath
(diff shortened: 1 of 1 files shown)


Luxonis-Brian:
@VanDavv - thanks for this! 

Everything makes sense, and is much better than the way I was trying to do it! I tested it out on the Raspberry Pi, and everything worked as expected. Going to test #69 now. 

LGTM.

PR #69, opened 2020-04-17
WIP get minimized timestamp diff pair

VanDavv:
Instead of checking the data_list for the most recent ones, we can try to find the minimum timestamp diff pair.

It's a potential improvement for #68


VanDavv:
This does not work due to the `data_list` containing just 0 or 1 packet at most

VanDavv:
I don't see any improvement (due to the very small size of `data_list`) that this PR would bring other than introduced in #68 , closing

PR #70, opened 2020-04-17
limit FPS on embedded platforms

### LINKED ISSUE #64: Check if calibration is being run on Pi, Set Framerate to 10FPS
The `why` of this is to reduce the latency on the Pi.  Since the Pi is doing some work in the background on calibration (reducing the stream resolution from 2x 1280x720 so you can see both easily on a 1920x1080 display), the Pi struggles to keep up and so the streams getting latent with high jitter in the latency.  This is especially painful when the streams aren't mirrored (see enhancement https://github.com/luxonis/depthai/issues/63), as you move the checkerboard in the wrong direction - and it takes a while for this to update on the screen - so by the time you realize you've moved the boards WAY in the wrong direction.

On faster hosts, they can easily keep up so this isn't necessary, but on the Pi it does start to struggle w/ 2x 1280x720 video feeds at 30FPS each and resizing them smaller for display fitting purposes.

So what would be great is to detect if calibration is being run on a Pi and automatically reduce the framerate to 10FPS.

To do this, it looks like (from [here](https://raspberrypi.stackexchange.com/questions/5100/detect-that-a-python-program-is-running-on-the-pi)) that you can just use `import os` and `os.uname()` and then pull out the nodename result and check if it is `raspberrypi`.

Then we could use the following specify the framerate of the streams coming from DepthAI (but change it for `left` and `right`:
https://github.com/luxonis/depthai/blob/91907b6854081e0e2998f5add74941c41209f1b3/depthai.py#L126

VanDavv:
This PR resolves #64 

I assume that machine specs starting with `arm` and `aarch64` are embedded systems, as listed [here](https://stackoverflow.com/questions/45125516/possible-values-for-uname-m)

### DIFF
--- a/calibrate.py
+++ b/calibrate.py
@@ -1,3 +1,4 @@
+import platform
 from contextlib import contextmanager
 
 import depthai
@@ -18,6 +19,8 @@ try:
     import cv2
 except ImportError:
     use_cv = False
+
+on_embedded = platform.machine().startswith('arm') or platform.machine().startswith('aarch64')
 
 
 def parse_args():
@@ -99,7 +102,9 @@ class Main:
     def __init__(self):
         self.args = vars(parse_args())
         self.config = {
-            'streams': ['left', 'right'],
+            'streams':
+                ['left', 'right'] if not on_embedded else
+                [{'name': 'left', "max_fps": 10.0}, {'name': 'right', "max_fps": 10.0}],
             'depth':
                 {
                     'calibration_file': consts.resource_paths.calib_fpath,


Luxonis-Brandon:
Oh ya that's a great assumption.  Will test this now.

Luxonis-Brandon:
Sorry got distracted.  This LGTM.  Latency is now way better on the Pi.

https://photos.app.goo.gl/Y6eWwFDGunJsDePY6

Luxonis-Brian:
LGTM. Thanks!

PR #71, opened 2020-04-17
Add ability to invert the horizontal and vertical axis

### LINKED ISSUE #63: Mirror the mono camera videos during calibration
The 'why' of this is if the video is not mirrored, you're instincts are to move the calibration checkerboard in the wrong direction.  It's like riding a bike with the handlebars reversed (it's really hard, see [here](https://www.youtube.com/watch?v=MFzDaBzBlL0)).

So this is why w/ FaceTime and other tools that show yourself, they mirror the video.  It's way more intuitive (in the FaceTime, they mirror it when they show it back to you, but then don't mirror it going out to the other party).

So we should do the same thing, mirror on display only, and pass the original frames-is to the calibration routine.

So I think the host (even the Pi) should be fast enough to do the flipping, but if not we can probably flip the streams in the Myriad X easily.

VanDavv:
Resolves #63 

Adds `-iv` and `-ih` (with long-names `--invert-vertical` and `--invert-horizontal` respectively) that flips the displayed image - for calibration, still the "un-flipped" frame is passed

### DIFF
--- a/calibrate.py
+++ b/calibrate.py
@@ -73,6 +73,10 @@ def parse_args():
                         help="Left/Right camera baseline in [cm]. Default: 9.0cm.")
     parser.add_argument("-w", "--no-swap-lr", dest="swap_lr", default=True, action="store_false",
                         help="Do not swap the Left and Right cameras. Default: True.")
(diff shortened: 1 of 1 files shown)


Luxonis-Brandon:
LGTM!  Thanks @VanDavv

PR #80, opened 2020-04-21
Update to MDK R11.6, OpenVINO 2020.2

alex-luxonis:


Luxonis-Brandon:
So far have run this on Mac OS X and it works fine.  Will test on Linux and RPi soon as well.

PR #82, opened 2020-04-24
Display EEPROM data, center L/R cameras

alex-luxonis:
1.
Data stored in EEPROM is printed when DepthAI starts, for example:
```
EEPROM data: valid (v3)
  Board name     : BW1098OBC
  Board rev      : R0M0E0
  HFOV L/R       : 71.86 deg
  HFOV RGB       : 68.7938 deg
  L-R   distance : 7.5 cm
  L-RGB distance : 3.75 cm
  L/R swapped    : yes
  L/R crop region: center
  Calibration homography:
    1.020374,   -0.009401,  -17.930922,
    0.014302,    1.018133,  -26.645380,
    0.000003,    0.000006,    1.000000,
```
or:
```
EEPROM data: invalid / unprogrammed
```

 2.
The full resolution for Left/Right mono cameras is 1280x800, and cropping is applied to obtain 1280x720. Until now the cropping was done using the top region. This PR adds support for center cropping (better aligned with RGB). Note: to be enabled, this requires recalibration, as otherwise the depth quality is degraded.

Implementation details: 
The `calibrate.py` script enables the `stereo_center_crop` flag, so centered images are captured by the cameras. After calibration processing, the flag is appended to the generated file `depthai.calib`. When the new calib file is used with `depthai.py`, by specifying the `-e/--store-eeprom` or `-o/--override-eeprom` option, the center cropping is enabled.

Backwards compatibility is maintained: if an existing older file `depthai.calib` (that doesn't contain the new flag) is written to EEPROM, the center cropping is disabled.

### DIFF
--- a/calibrate.py
+++ b/calibrate.py
@@ -144,6 +144,7 @@ class Main:
                     'swap_left_and_right_cameras': self.args['swap_lr'],
                     'left_fov_deg':  self.args['field_of_view'],
                     'left_to_right_distance_cm': self.args['baseline'],
+                    'override_eeprom': True,
                 }
         }
         if self.args['board']:
--- a/calibrate.py
+++ b/calibrate.py
@@ -145,6 +145,7 @@ class Main:
                     'left_fov_deg':  self.args['field_of_view'],
                     'left_to_right_distance_cm': self.args['baseline'],
                     'override_eeprom': True,
+                    'stereo_center_crop': True,
                 }
         }
         if self.args['board']:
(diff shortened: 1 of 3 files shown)


Luxonis-Brian:
Awesome @alex-luxonis. Thanks. Will test out tomorrow morning.

Luxonis-Brandon:
Wow I love this display!   I have to crank out two proposals tonight otherwise I would test this now.  

This will be so great to have!

Luxonis-Brandon:
Trying this out now!

Luxonis-Brandon:
So I did some more testing on this, this time on the Raspberry Pi Compute Module and I did see a segmentation fault at the end calibrate.py on  the closure of the XLINK streams.  

So @szabi-luxonis and @alex-luxonis I'm wondering if there is something here maybe WRT the upgrade to R10.15 or something...  as this is the same crash I see on Mac OS X.  I'm wondering if it just happens more often on Mac OS X (on Master).  Anyway, will make a new issue about it as I don't think this has to do w/ this PR.

So the calibrations is successful, see below:
image

image
image
image

szabi-luxonis:
LGTM

PR #84, opened 2020-04-28
Sanitize NN output (NaN, Infinite); handle corrupted frames, LRU frame queue replace policy

szabi-luxonis:
Add checks in NN decode to handle floating point exceptions.
Discard corrupted frames.
Implement LRU replace policy for frame queue, replace the oldest frame in the queue in case of overflow.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,15 +6,19 @@ __Documentation is available at [https://docs.luxonis.com](https://docs.luxonis.
 
 ## Python modules
 
+DepthAI requites [numpy](https://numpy.org/) and [opencv-python](https://pypi.org/project/opencv-python/). To get the versions of these packages you need for DepthAI, use pip: `pip install -r requirements.txt`
+
 Files with `.so` extension are the python modules:  
 - `depthai.cpython-36m-x86_64-linux-gnu.so` built for Ubuntu 18.04 & Python 3.6  
-- `depthai.cpython-37m-arm-linux-gnueabihf.so` built for Raspbian 10 & Python 3.7  
+- `depthai.cpython-37m-arm-linux-gnueabihf.so` built for Raspbian 10 & Python 3.7
 
 For supporting other platforms, there is an option to build the python lib from sources by grabbing the [depthai-api](https://github.com/luxonis/depthai-api) submodule:
 
     git submodule update --init
     ./depthai-api/install_dependencies.sh # Only required in first build on a given system
     ./depthai-api/build_py_module.sh
+
+When updating DepthAI on these platforms it is often necessary to run `./depthai-api/build_py_module.sh --clean` in order to build a new version of the depthai-api module for your chosen platform. 
 
 ## Examples
 
--- a/calibrate.py
+++ b/calibrate.py
@@ -126,8 +126,7 @@ class Main:
         self.args = vars(parse_args())
         self.config = {
             'streams':
-                ['left', 'right'] if not on_embedded else
-                [{'name': 'left', "max_fps": 10.0}, {'name': 'right', "max_fps": 10.0}],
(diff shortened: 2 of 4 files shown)


Luxonis-Brandon:
Thanks!  Chris and Lukasz were just needing this!

Luxonis-Brandon:
Good idea; I like that - I actually hadn’t checked the files changed yet and had assumed (incorrectly) that this was being added to the C++ backend.

szabi-luxonis:
> Would it be possible to move some of these validations in the C++ lib backend?
> In the idea of keeping the python script clean...

TensorEntry validations were moved into library

Luxonis-Brandon:
This works find for me on OS X.

Luxonis-Brandon:
Works fine on BW1097 also.   

Going to let it run over night pointed at its own monitor (so it's constantly detecting at least something).

Luxonis-Brandon:
Mention in slack already but reproducing here for posterity:

Brandon:

> So @szabi-luxonis  with the sanitize PR it successfully prevented the NaN and other errors which would usually cause the watchdog to have to trigger.
> 
> - And ran for at least 6 hours if not a lot more.
> - I think exposed a new one that occurs much less often probably, above ^
> - Manually re-running it after this it runs fine.
> - Note that the watchdog did not catch this one...

Szabi:

> I was able to reproduce this issue artificially on PC, I will push a possible fix in a few minutes.

Luxonis-Brandon:
Thanks for the update/fix for this.  Going to stop the Pi from running and get it doing a 24-hour run again.

And also test this latest on Mac and Ubuntu real quick.

Luxonis-Brandon:
This is now running on the BW1097 (started 33 minutes ago), on Mac (just now) and Ubuntu (also just now).  

Will let them all run for a long time and circle back.

Luxonis-Brandon:
So I can confirm @szabi-luxonis that the error that occurred last time which caused the host code to stop/exit (the `too many indices for array` issue) did occur on this run and your fix successfully prevented it from causing a problem.  

image
image

So I think this one is good to go!

szabi-luxonis:
Nice.
I'm running it also on RPI, since I pushed the fix, I will check tomorrow morning, and if it's fine I will merge it.

Luxonis-Brandon:
So @szabi-luxonis I'm seeing a new one on this PR: The neural output never occurs.  I'm wondering if something in sanitization is making it so all future results get blocked or similar.

I've started it on BW1097 (`python3 test.py`) and seen no network results 3 times in a row now.

Luxonis-Brandon:
After hitting reset button on Myriad X it started working again.

Luxonis-Brandon:
And I had run `python3 calibrate.py` prior to running `python3 test.py`.  So it could be some state-info stays after or something from calibrate.py which causes this problem.

Luxonis-Brandon:
Haven’t seen that boot up issue again. And have been running this on BW1097 for 16+ hours now.

Luxonis-Brandon:
So I meant to circle back that I tested this for a long time over the weekend.  My Raspberry Pi eventually hard-froze (mouse won't even move).  It seems to have run out of RAM, as evidenced by the HD LED being permanently on (see below):
image

So I tried going back to terminal only to get the system responsive (ctrl-alt-f6 IIRC), but it wouldn't budge and when I came back an hour or so later it had moved to terminal only, but wasn't usable.

So I had seen this exact behavior before on this same Pi when not doing anything with DepthAI, and it was immediately subsequent to TeamViewer being installed on this Pi... so the possibility exists that this has nothing to do with DepthAI.  

I've since removed teamviewer and I'm doing another long-term run.  It's ~36 hours in so far and fine.

Issue #83, opened 2020-04-27
Segmentation Fault with calibrate.py on BW1097


Luxonis-Brandon:
When running calibrate.py on BW1097 the system seg faults after having stepped through checking out the stereo pairs for visualization.  

This seems to be similar to the XLink issue that @VanDavv was seeing when running DepthAI on Raspberry Pi:

```
python3 calibrate.py -s 2.2 -ih
Using Custom Calibration File: depthai.calib
Using Arguments= {'count': 1, 'square_size_cm': 2.2, 'image_op': 'modify', 'mode': ['capture', 'process'], 'config_overwrite': None, 'board': None, 'field_of_view': 71.86, 'baseline': 9.0, 'swap_lr': True, 'dev_debug': None, 'invert_v': False, 'invert_h': True}
Starting image capture. Press the [ESC] key to abort.
Will take 13 total images, 1 per each polygon.
XLink initialized.
Sending device firmware "cmd_file": /home/pi/Desktop/depthai/depthai.cmd
Successfully connected to device.
Loading config file
Attempting to open stream config_d2h
Successfully opened stream config_d2h with ID #0!
Closing stream config_d2h: ...
Closing stream config_d2h: DONE.
WARNING: Version (dev) does not match (c722ebde932d6627463321816a5654b5be6069e1 & b8cebab238c71263c80087051a70c8165f95d374)
EEPROM data: valid (v3)
watchdog started 1000
  Board name     : BW1097
  Board rev      : R1M1E2
  HFOV L/R       : 71.86 deg
  HFOV RGB       : 68.7938 deg
  L-R   distance : 9 cm
  L-RGB distance : 2 cm
  L/R swapped    : yes
  L/R crop region: center
  Calibration homography:
    1.015737,    0.005940,  -21.193869,
   -0.002906,    1.014818,  -16.099894,
    0.000003,    0.000003,    1.000000,
CNN configurations read: /home/pi/Desktop/depthai/resources/nn/mobilenet-ssd/mobilenet-ssd.json
config_h2d json:
[...]

Issue #78, opened 2020-04-20
Calibration Fails on Mac OS X with Segmentation Fault


Luxonis-Brandon:
When running calibration on Mac OS X it seems to SegFault on or around saving the 13th image.  

Tested on https://github.com/luxonis/depthai/pull/66/ and also Master (49aad6662d18850a0544a209eba9fcf3ab624da3).

```
git submodule update --init
./depthai-api/build_py_module.sh
grep: /proc/meminfo: No such file or directory
MemAvailable:  MB
Unable to get MemAvailable
[BUILD depthai-api/host/py_module/build] make -j4
[ 14%] Built target nlohmann_json_schema_validator
[100%] Built target depthai
[BUILD depthai-api/host/py_module/build] cp depthai.cpython-37m-darwin.so ../../../../

python3 calibrate.py -ih
No calibration file. Using Calibration Defaults.
Using Arguments= {'count': 1, 'square_size_cm': 2.5, 'image_op': 'modify', 'mode': ['capture', 'process'], 'config_overwrite': None, 'field_of_view': 71.86, 'baseline': 9.0, 'swap_lr': True, 'dev_debug': None, 'invert_v': False, 'invert_h': True}
Starting image capture. Press the [ESC] key to abort.
Will take 13 total images, 1 per each polygon.
XLink initialized.
Sending device firmware "cmd_file": /Users/leeroy/depthai-python-extras/depthai.cmd
Successfully connected to device.
Loading config file
Attempting to open stream config_d2h
watchdog started 1000
Successfully opened stream config_d2h with ID #0!
Closing stream config_d2h: ...
Closing stream config_d2h: DONE.
WARNING: Version (dev) does not match (unknown & 62b4db70b327f51f6066af9620187ee4ae3878b8)
CNN configurations read: /Users/leeroy/depthai-python-extras/resources/nn/object_detection_4shave/object_detection.json
depthai: Calibration file is not specified, will use default setting;
config_h2d json:
[...]

PR #86, opened 2020-04-28
Watchdog and segfault fix

szabi-luxonis:
USB watchdog was broken in commit `7ca02816b5176f685ff38d96bea07cafc0ece5f3` due to some changes in XLink code on newer openvino 2020.1 repo. 
On program termination segfault was raised, due to mishandling of a shared pointer between python and CPP.
This PR fixes both issues. Issue #83  and #78.

PR depends on #88, #88 should be merged first.

### DIFF
--- a/depthai.py
+++ b/depthai.py
@@ -81,6 +81,8 @@ def parse_args():
                         Format: stream_name or stream_name,max_fps \
                         Example: -s metaout previewout \
                         Example: -s metaout previewout,10 depth_sipp,10")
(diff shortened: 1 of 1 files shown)


szabi-luxonis:
> `deinit_device()` takes about 4..6 seconds (measured after `q` is pressed).
> Is there some way we could improve this?

We can improve it, by reducing watchdog polling rate instead of 3000 ->30*100 ms.
Or the dirty way, calling os._exit(0), letting the os to handle deallocation almost instantly.

PR #99, opened 2020-05-08
Model compiler

szabi-luxonis:
From model_compiler folder run:
./download_and_compile.sh 
The following command will compile the models from cloud:
./download_and_compile.sh CLOUD_COMPILE


GotG:
it works with repo default model folders in `resources/nn`
if user adds a custom folder in `nn`  it does not work (since the model name is not available in the open model zoo)

Issue #98, opened 2020-05-07
calibrate.py sometimes does nothing when hitting space bar


Luxonis-Brandon:
On the BW1097 sometimes calibrate.py does nothing at all when hitting space bar.  No green display of it working, no error display, etc.  It just sits there for a while.

We should correct this.  If the system is stuck processing, we should display this so that the user knows what's going on.

Luxonis-Brandon:
And to elaborate on this... it doesn't even print anything to the terminal.

Luxonis-Brandon:
So @szabi-luxonis I think I figured this one out... calibrate.py is doing 10FPS on the grayscale cameras so on USB frames are being dropped (20/s) for each camera and we're not making an effort to choose which frames are dropped, so the left/right that make it (1 out of 3) don't necessarily end up being of the same time stamp...   And calibrate.py is checking to make sure the left/right frames have a timestamp that is is within 1ms of eachother.  So when this fails, the script just sits there.

Luxonis-Brandon:
So I jsut confirmed this must the issue by just disabling the framerate limit here: 
https://github.com/luxonis/depthai/blob/0611eb701fffbcd743644e4f0d2c7b92e60b7cd9/calibrate.py#L130

PR #100, opened 2020-05-08
Update calibrate.py

Luxonis-Brandon:
Limiting the framerate results in getting frames that are sometimes not in sync.  

So I removed the 10FPS option on embedded for now until we figure out a way to not cause the frames to sometimes be out of sync.

Seems to solve #98

### DIFF
--- a/calibrate.py
+++ b/calibrate.py
@@ -126,8 +126,7 @@ class Main:
         self.args = vars(parse_args())
         self.config = {
             'streams':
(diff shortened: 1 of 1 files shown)


szabi-luxonis:
lgtm as temporary solution

Luxonis-Brandon:
Just going to merge this as calibrate.py is annoying to use w/out this workaround.

PR #95, opened 2020-05-06
UI/UX Improvements in visuals and dev cycle

philnelson:
So far this PR does the following:

* Implements helper functions to draw nice labels and bounding boxes.
* Implements new display for mobilenet labeled objects which uses the aforementioned helper functions, it looks like this:

Screen Shot 2020-05-06 at 5 10 36 PM

* Add note about needing to use --clean on build commands after updates.
* Improves clarity of a few code comments
* Adds a python requirements file
* Resizes and repositions the FPS counter

PR #101, opened 2020-05-08
Revert "UI/UX Improvements in visuals and dev cycle"

Luxonis-Brandon:
Reverts luxonis/depthai#95

This does not work with `-bb` on `depth_sipp`:
`python3 test.py -s metaout previewout depth_sipp,5 -bb`

image

So just reverting for now until we get it compatible there.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,8 +6,6 @@ __Documentation is available at [https://docs.luxonis.com](https://docs.luxonis.
 
 ## Python modules
 
-DepthAI requites [numpy](https://numpy.org/) and [opencv-python](https://pypi.org/project/opencv-python/). To get the versions of these packages you need for DepthAI, use pip: `pip install -r requirements.txt`
-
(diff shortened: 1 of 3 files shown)


Luxonis-Brandon:
image
Here's after the revert.  Going to just merge this real quick as we are needing this overlay for a demo.

philnelson:
Huh- why are we defining the bounding box differently when using depth_sipp? Should've tested, my bad.

Luxonis-Brandon:
You’re good man!  I knew about this feature and totally forgot to mention it test it etc. even though I fully knew about it.  

So the why is with pulling depth to calculate xyz information of the object there’s this variable padding_factor which lets you decide what portion of the object detector bounding box is used to calculate object depth.  And it’s useful to display that inner box so you can see if it’s working correctly for your object of interest. 

I totally forgot about this when discussing the changes to the UI.  My bad.

PR #102, opened 2020-05-09
When max_fps is requested, drop frames based on sequence number 

### LINKED ISSUE #98: calibrate.py sometimes does nothing when hitting space bar

szabi-luxonis:
When max_fps is requested, drop frames based on sequence number rather than arrival time.
Fixes #98

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,15 +6,19 @@ __Documentation is available at [https://docs.luxonis.com](https://docs.luxonis.
 
 ## Python modules
 
+DepthAI requites [numpy](https://numpy.org/) and [opencv-python](https://pypi.org/project/opencv-python/). To get the versions of these packages you need for DepthAI, use pip: `pip install -r requirements.txt`
+
 Files with `.so` extension are the python modules:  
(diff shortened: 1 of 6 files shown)


Luxonis-Brandon:
Ah sweet thanks!  

So as you mention this should address the issue we’re seeing with calibrate.py when limiting the framerate.

szabi-luxonis:
Yes, just reverted commit that removed limiting on embedded, it's back to 10 fps for calibrate.py.
Also, max_fps is consistent now, if you request 20 fps, you receive 20.

Luxonis-Brandon:
Oh double-sweet thanks!

szabi-luxonis:
Combining with : https://github.com/luxonis/depthai-api/pull/18/commits/dbb699672a966b29dd45bb582b6918dc25dc6879
on another PR, will also get rid of data queue full messages.

PR #104, opened 2020-05-12
Python3 shebang line; executable permission for python scripts

szabi-luxonis:
To avoid problems with wrong environment/python version this PR adds possibility to run scripts with any of the following commands:
python depthai.py
python3 depthai.py
./depthai.py
Tested on Linux.

### DIFF
--- a/calibrate.py
+++ b/calibrate.py
@@ -1,3 +1,5 @@
+#!/usr/bin/env python3
+
 import platform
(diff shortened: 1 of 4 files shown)


Luxonis-Brandon:
Oh good idea thanks!

alex-luxonis:
Should we add the change to `calibrate*.py` as well?

PR #106, opened 2020-05-13
Add cli utils

rhollen:
This pull request changes and adds no functionality. It is purely to improve organization of code by breaking out like functions into a separate module. It also introduces some simple tests to be expanded on as the project grows. I do not have hardware on hand to test that the system runs as expected, but outputs to each function should remain the same and is tested.

Happy to update or add additional testing coverage.

### DIFF
--- a/depthai.py
+++ b/depthai.py
@@ -14,7 +14,7 @@ import depthai
 
 import consts.resource_paths
 from depthai_helpers import utils
-from depthai_helpers.cli_utils import parse_args, PrintColors
+from depthai_helpers.cli_utils import cli_print, parse_args, PrintColors
 
 
 def decode_mobilenet_ssd(nnet_packet):
@@ -174,7 +174,7 @@ if args['config_overwrite']:
 print("Using Arguments=",args)
 
 if args['force_usb2']:
-    print(PrintColors.WARNING + "FORCE USB2 MODE" + PrintColors.ENDC)
+    cli_print("FORCE USB2 MODE", PrintColors.WARNING)
     cmd_file = consts.resource_paths.device_usb2_cmd_fpath
 else:
     cmd_file = consts.resource_paths.device_cmd_fpath
@@ -216,11 +216,11 @@ if args['cnn_model']:
 blob_file_path = Path(blob_file)
 blob_file_config_path = Path(blob_file_config)
 if not blob_file_path.exists():
-    print(PrintColors.WARNING + "\nWARNING: NN blob not found in: " + blob_file + PrintColors.ENDC)
+    cli_print("\nWARNING: NN blob not found in: " + blob_file, PrintColors.WARNING)
     os._exit(1)
 
 if not blob_file_config_path.exists():
-    print(PrintColors.WARNING + "\nWARNING: NN json not found in: " + blob_file_config + PrintColors.ENDC)
+    cli_print("\nWARNING: NN json not found in: " + blob_file_config, PrintColors.WARNING)
     os._exit(1)
 
 with open(blob_file_config) as f:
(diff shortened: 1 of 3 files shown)


Luxonis-Brandon:
Woh, @rhollen !  Thanks for the PR!  Been a crazy long time man, I think 14 years?

We're more than happy to send you a unit to play with to say thanks.  Will reach out offline.

rhollen:
@Luxonis-Brandon -- it has been a long time! Thanks for the offer, would be fun to actually use the hw :)

@ szabi-luxonis -- please let me know how to best contribute. I haven't been involved much in open source contributions, so please assume ignorance if I am doing anything wrong!

Luxonis-Brandon:
Thanks @rhollen !  Discussing internally it looks great!  

I'll test it as well on Mac/Ubuntu/Raspbian with actual DepthAI hardware to make sure there isn't some unforeseen thing here.  Doesn't seem like there would be though.

And we'll get your unit shipped out to you today probably.  :-)

Luxonis-Brandon:
OK finally got around to testing this w/ actual hardware.  

Ran on Mac:
- with BW1093 w/ `python3 test.py` and it worked fine.
image

- with BW1098OBC w/ `python3 test.py -s metaout previewout depth_sipp,5 -bb -ff -cnn face-detection-retail-0004` and it worked fine.  
image
image

Going to run it on BW1097 and w/ BW1098 + Ubuntu soon here.

philnelson:
Thanks for the refactor. For the most part seems good, but a couple of thoughts/questions:

* One thing that is a bit odd to me is having the argparser info for a script (depth.py) in a totally separate file. I'd expect the arg refs for `depthai.py` to be in `depthai.py`. Clean up is good, but I wonder if that is a bit too clean so to speak.

* It also seems to make the assumption that the tests and the depth.py file will always have the same args, but I'm not sure that's the case. This is more a design issue/question.

* I'm not sure I agree with creating our own `print` override just to add colors with `cli_print` but that might just be me being weird.

Having some tests is nice!

Luxonis-Brandon:
Finally tested this on Raspberry Pi and Ubuntu.  All works fine.  

Ubuntu:
image

Luxonis-Brandon:
> Thanks for the refactor. For the most part seems good, but a couple of thoughts/questions:
> 
> One thing that is a bit odd to me is having the argparser info for a script (depth.py) in a totally separate file. I'd expect the arg refs for depthai.py to be in depthai.py. Clean up is good, but I wonder if that is a bit too clean so to speak.
> 
> It also seems to make the assumption that the tests and the depth.py file will always have the same args, but I'm not sure that's the case. This is more a design issue/question.
> 
> I'm not sure I agree with creating our own print override just to add colors with cli_print but that might just be me being weird.
> 
> Having some tests is nice!

So @philnelson discussed this offline.  I'm a fan of breaking out the CLI functions into their own script so that a user who doesn't care about the CLI interface can see the rest of the code/how-to w/out having to scroll past it all.

So we're good to go w/ that @rhollen

rhollen:
Thanks @philnelson!  -- I'll answer your comments inline

> One thing that is a bit odd to me is having the argparser info for a script (depth.py) in a totally separate file. I'd expect the arg refs for depthai.py to be in depthai.py. Clean up is good, but I wonder if that is a bit too clean so to speak.

I think this is a pretty standard architecture, to break up like functions into modules. It helps organize the code as it grows and makes it easier to find what you need. I noticed that the `depthai.py` module was getting very large and was hard to understand coming from the outside. As this project continues to grow, it will be increasingly harder to contribute. 

> It also seems to make the assumption that the tests and the depth.py file will always have the same args, but I'm not sure that's the case. This is more a design issue/question.

I had no ability to test this at the time. It was mostly to add _some_ level of testing and anyone / everyone should add more cases to the tests. Hopefully these tests can become a bit more automatic such that testing changes becomes less manual, but this is a very tiny, baby step in that direction.

> I'm not sure I agree with creating our own print override just to add colors with cli_print but that might just be me being weird.

It's not an override (I've seen that done, where python built in `print()` gets overridden -- bad!), it's just adding a function to do the formatting. It is a bit "cute", but I thought it made the code read a bit cleaner. It's also less error prone and less repeated code. I think it's an improvement,  not a required change, but you could say that about the whole pull request!

Thanks for the review!

VanDavv:
@rhollen Suggestions from my side. 

First, I would rename `cli_utils` to just `cli`. It's cleaner and specifies that these are not just utilitary functions (e.x. parse_args is a core functionality I think)

Also, Instead of doing `cli_print` and passing two arguments, I would provide methods to `cli` class in a manner similar to `logging` python module, so we'd have `cli.warning("sometext")`, and also provide `cli.print("sometext", PrintColor.RED)` if someone would like to do something custom.

I love the tests, also felt a need for them and it's great that you brought them up @rhollen, thanks!

VanDavv:
@rhollen on second thought, I'll just prepare a separate PR with these changes, don't want to hold these up - as they're definitely needed and much appreciated your contribution!

rhollen:
@VanDavv -- thanks for the review! I do like your suggestions. I named it `cli_utils` to match the `calibration_utils` naming, but we can rename. Simpler is better.

As for the other change request, I also like it. I went with the function because I thought it fit in better with most the other code I saw and it was the smallest change I could think of. It will be a bit more code to write it as you've suggested, but it is nice to have it look like the logging module. Feel free to update, or I can open up another PR. Please let me know. I do think a separate pull request is better here.

VanDavv:
I think the same, let's merge this one. Thanks @rhollen ! Looking forward to work with you on next PRs!

PR #109, opened 2020-05-15
Show synchronized left and right captured frames in a single window

szabi-luxonis:
Show synchronized left and right captured frames in a single window


szabi-luxonis:
Can be changed to horizontal stacking at https://github.com/luxonis/depthai/pull/109/files#diff-8c59fe2157d026536287a0016f5d7d07R311

PR #110, opened 2020-05-17
fix std::system_error; set usb_loglevel to 1

szabi-luxonis:
Fix std::system_error on boot error + set usb_loglevel to 1 (print errors on boot)


Luxonis-Brandon:
- Works fine on Mac.
- Tested on Raspbian, works fine too.
- Tested on Ubuntu and works well too.

PR #114, opened 2020-05-25
Trigger autofocus and change mode feature

themarpe:
Example usage is shown in depthai.py
Press key: 
 - 'f' to trigger autofocus
 - '1' to change autofocus mode to auto
 - '2' to change autofocus mode to continuous video

Other autofocus modes are also available (check depthai.AutofocusMode.__members__) but not yet tested.


Luxonis-Brandon:
I'd say this is good to merge!  :-)

PR #117, opened 2020-05-27
Add back time.monotonic import

alex-luxonis:
`from time import monotonic` got somehow deleted when PR https://github.com/luxonis/depthai/pull/106 was merged.
Adding it back, to fix the crash in the python script.


Luxonis-Brandon:
Works for me.  
image

PR #121, opened 2020-06-01
Update depthai-api (python3_exact_version)

alex-luxonis:
See PR https://github.com/luxonis/depthai-api/pull/28

> With multiple `python3` versions installed (side by side, or within Conda environments), 
> CMake's `find_package` always returned the latest (e.g. even if we specify 3.6, if 3.7 is found, it is used instead).
> With an older python3 interpreter enabled as default, this was causing a wrong lib to be built, not the one expected by the selected python3 interpreter.
> 
> However, `find_package` has an option `EXACT` that would solve the above issue, but we don't want to hardcode a specific version.
> We query the version from the python3 interpreter and pass it to `find_package`.


Luxonis-Brandon:
Tested on Mac and works fine.  I'd say this is good to go.

szabi-luxonis:
lgtm

PR #128, opened 2020-06-06
Keep NN packet source alive while it's used to avoid dangling ptr/reference

szabi-luxonis:
In the code we parse tensors, which are stored in `TensorEntryContainer` and store them in a list `prev_entries`.
`TensorEntryContainer` (nnet_packet) went out of scope after parsing 
(after  `for _, nnet_packet in enumerate(nnet_packets): `), the reference was lost to it and it was deallocated, but the tensors parsed were pointing to the `TensorEntryContainer` object allocated on heap (dangling pointer). Any object allocated after TensorEntryContainer deallocation had a chance that it was allocated in the same memory space where `TensorEntryContainer` was before, thus corrupting tensors which were accessed through `prev_entries`.

This issue is clearly reproducible in `https://github.com/luxonis/depthai/tree/nn_left_right_BUGFIX` where the NN s for left camera are corrupted. They were more susceptible to corruption than right, because FPS on left camera is lower, so NN packets for right camera were allocated more often, corrupting the deallocated `TensorEntryContainer` for left.

The fix saves the `nnet_source` between 2 NNet packets, thus holding the `shared_ptr`, to avoid deallocation. Python bindings are fine.

### DIFF
--- a/depthai.py
+++ b/depthai.py
@@ -402,7 +402,10 @@ for s in stream_names:
     frame_count[s] = 0
     frame_count_prev[s] = 0
 
-entries_prev = []
+nnet_prev = {}
+nnet_prev["entries_prev"] = []
+nnet_prev["nnet_source"] = []
+
 tracklets = None
 
 process_watchdog_timeout=10 #seconds
@@ -429,7 +432,8 @@ while True:
             os._exit(10)
 
     for _, nnet_packet in enumerate(nnet_packets):
-        entries_prev = decode_nn(nnet_packet)
+        nnet_prev["nnet_source"] = nnet_packet
+        nnet_prev["entries_prev"] = decode_nn(nnet_packet)
 
     for packet in data_packets:
         if packet.stream_name not in stream_names:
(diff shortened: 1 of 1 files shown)


depthai.py
```diff
@@ -402,7 +402,10 @@ def show_landmarks_recognition(entries_prev, frame):
     frame_count[s] = 0
     frame_count_prev[s] = 0
 
-entries_prev = []
+nnet_prev = {}
+nnet_prev["entries_prev"] = []
+nnet_prev["nnet_source"] = []
```
VanDavv, line 407: `nnet_source` seems to be unused, maybe we can move back to `entries_prev = []`?

depthai.py
```diff
@@ -429,7 +432,8 @@ def reset_process_wd():
             os._exit(10)
 
     for _, nnet_packet in enumerate(nnet_packets):
-        entries_prev = decode_nn(nnet_packet)
+        nnet_prev["nnet_source"] = nnet_packet
```
szabi-luxonis, line 435: @VanDavv  nnet_source is used here, holding the shared_ptr from cpp backend from going out of scope and deallocating the object

depthai.py
szabi-luxonis: see my comment below

depthai.py
VanDavv: Ok, got it, so we need to save it somewhere so it won't be deleted - thanks!

depthai.py
szabi-luxonis: The logic is that `entries_prev` and `nnet_source` are tied together, that's why I created a dictionary instead 2 lists.

Issue #129, opened 2020-06-08
Application crashes when running for a long period of time


VanDavv:
## Overview

I've been able to reproduce and verify a couple times that running the test.py for a long time (1-24 hrs), and using more streams than just `previewout` and `metaout`, results in the application crashes (mostly segfaults, but there are some other errors too, including Python type mappings errors).

## Errors
Most recent configurations ran:

- `python3.7 depthai.py --streams previewout metaout --video /dev/null` crashed with the following backtrace:

    ```
    Error while getting packet data: Unable to convert function return value to a Python type! The signature was
        (self: depthai.DataPacket) -> array
    Thread 1 "python3.7" received signal SIGFPE, Arithmetic exception.
    raise (sig=8) at ../sysdeps/unix/sysv/linux/raise.c:50
    50  ../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
    (gdb) backtrace
    #0  0x76f378ec in raise (sig=8) at ../sysdeps/unix/sysv/linux/raise.c:50
    #1  0x76ca51f8 in __aeabi_ldiv0 ()
        at /lib/arm-linux-gnueabihf/libgcc_s.so.1
    #2  0x71d84c40 in HostDataPacket::getPythonNumpyArray() ()
        at /home/pi/Desktop/depthai/depthai.cpython-37m-arm-linux-gnueabihf.so
    #3  0x71db8640 in pybind11::cpp_function::cpp_function<pybind11::array*, HostDataPacket, , pybind11::name, pybind11::is_method, pybind11::sibling, pybind11::return_value_policy>(pybind11::array* (HostDataPacket::*)(), pybind11::name const&, pybind11::is_method const&, pybind11::sibling const&, pybind11::return_value_policy const&)::{lambda(HostDataPacket*)#1}::operator()(HostDataPacket*) const ()
        at /home/pi/Desktop/depthai/depthai.cpython-37m-arm-linux-gnueabihf.so
[...]

PR #130, opened 2020-06-10
Guard spsc queue push with mutex since it's not thread safe

szabi-luxonis:
SPSC (single producer-single consumer) queue [documentation](https://www.boost.org/doc/libs/1_53_0/doc/html/boost/lockfree/spsc_queue.html
) about queue `push` function:
`Requires: | only one thread is allowed to push data to the spsc_queue`
We had multiple threads calling `HostPipeline::onNewData` once USB packet arrived, queue push wasn't protected, so the data got released in between multiple threads pushing into queue and main python thread (consumer) consuming the queue.

This issue can be reproduced easily on commit 95d2b3450596d08624cf7543dc52c702a922502c
w/ options : `./depthai.py -s previewout left right metaout -v /dev/null`
where each thread is spamming the queue -artificially- every 1 ms to reproduce the clash.
When the issue happens in the terminal it will be printed (most likely) : `Invalid packet stream name!deallocated`
but technically it's subject to undefined behavior due to corruption.

Protecting the queue in both consumer and producer defies the purpose of non locking queue, the purpose of this PR is not replacing the current queue implementation with a locking one, but to solve the current issue in the current context.


szabi-luxonis:
Fixes issue #129 .

PR #139, opened 2020-06-30
Adding AprilTags

jonngai:


szabi-luxonis:
Closed, since it depends on master which will be deleted.

PR #156, opened 2020-07-20
Merge latest master into develop

iamsg-luxonis:
Merge latest master code into develop. Including NN on left right, memory optimizations etc.

Build as :
python3 -m pip install -r requirements.txt


Luxonis-Brandon:
This is huge, thanks!

PR #157, opened 2020-07-21
Sync NN meta with depth(x,y,z), sync previewout & metaout

alex-luxonis:
- Sync NN metadata with depth frames for X,Y,Z calculation, based on camera capture timestamps.
As RGB and Mono cameras are not synchronized, another problem we saw was a slow drift, 1 frame every 80 seconds at 30fps. To circumvent this, a software-sync method is implemented: based on RGB-Mono timestamp diff, a control loop is adjusting the RGB frame interval, in small amounts by up to ± 0.2ms. During startup, it may take few seconds (up to 10), until RGB and Mono get in close sync together (this could be further improved). Until then, then max timestamp diff between RGB-Mono is no larger than half frametime (that is 16.6ms at 30fps). 
Note: for this mechanism to work well, if the RGB and Mono FPS are modified from the default of 30 (using `-rgbf` and `-monof` arguments), they should be set to the same value.

- Add an option to improve the synchronization of NN metadata overlay on the previewout stream, by aligning the packets sent from device:
```
  -sync, --sync-video-meta
                        Synchronize 'previewout' and 'metaout' streams

```
Note: for this to work well, the NN FPS should not be slower than camera FPS. To improve on this (for some NN models), run NN with 2 NCEs and more Shaves, or reduce RGB and Mono FPS, for example:
./depthai.py -nce 2 -sh 10 -cmx 10 -sync
./depthai.py -rgbf 24 -monof 24 -sync

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,14 @@ __Documentation is available at [https://docs.luxonis.com](https://docs.luxonis.
 
 ## Python modules
 
-DepthAI requites [numpy](https://numpy.org/) and [opencv-python](https://pypi.org/project/opencv-python/). To get the versions of these packages you need for DepthAI, use pip: `pip3 install -r requirements.txt`
+DepthAI requires [numpy](https://numpy.org/) and [opencv-python](https://pypi.org/project/opencv-python/). To get the versions of these packages you need for DepthAI, use pip: `python3 -m pip install -r requirements.txt`
+
+Optional:
+For command line autocomplete when pressing TAB, only bash interpreter supported now:
+Add to .bashrc:
+`echo 'eval "$(register-python-argcomplete depthai.py)"' >> ~/.bashrc`
+
+If you use any other interpreter: https://kislyuk.github.io/argcomplete/
 
 Files with `.so` extension are the python modules:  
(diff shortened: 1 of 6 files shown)


szabi-luxonis:
LGTM

VanDavv:
Tested with 8 different networks and they all performed great - I've even got feeling that some tracker is attached to my face with `face-detection-retail-0004` as both network and preview were running with 30 FPS and the sync was visible.
With slower network - `face-detection-adas-0001`, got 8 FPS and the preview also went down to 8 FPS and both results and image were consistent.

LGTM, great to have this @alex-luxonis !

Luxonis-Brandon:
Thanks @VanDavv for testing and sorry @alex-luxonis that I've been so far behind on testing.  Trying this now.

Luxonis-Brandon:
This works insanely well @alex-luxonis !  I agree with Lukasz comment of it feeling glued to you.  I ran `python3 test.py -nce 2 -sh 10 -cmx 10 -sync` and it was fantastic how well synced and fast everything was!

See the quick video below:
AI](https://i.imgur.com/nXVZkeC.jpg)](https://www.youtube.com/watch?v=BCa5z1ssxxo "DepthAI on Mac")
https://youtu.be/BCa5z1ssxxo
https://i.imgur.com/nXVZkeC.jpg

PR #158, opened 2020-07-21
Fix error when previewout is not enabled : -s metaout depth_sipp

szabi-luxonis:
fix error when previewout is not enabled : -s metaout depth_sipp


Luxonis-Brandon:
Thanks!  Testing now on Mac as well.

Luxonis-Brandon:
image
Yep, works great now, thanks!  I think good to merge.

PR #148, opened 2020-07-08
updated LUT to improve depth_color_h visualization

Luxonis-Steven:
updated disparity LUT in disparity_luts.hpp to improve this: https://github.com/luxonis/depthai/issues/90

PR #162, opened 2020-07-22
Python command line autocomplete

szabi-luxonis:
Add support for argument autocomplete on TAB pressing, like any other app.
`python3 -m pip install -r requirements.txt`
`echo 'eval "$(register-python-argcomplete depthai.py)"' >> ~/.bashrc`

Only bash is supported for now.
On macos run depthai from bash: from terminal: `bash` then follow instructions.

When pressing tab it will list all available arguments, and all available options for that argument, for example:
`./depthai.py -cnn` TAB will list all supported CNN models.
`./depthai.py -cnn mobi` TAB will autocomplete to mobilenet-ssd

### DIFF
--- a/depthai_helpers/cli_utils.py
+++ b/depthai_helpers/cli_utils.py
@@ -1,8 +1,39 @@
 #!/usr/bin/env python3
 
 import argparse
-from enum import Enum
+try:
+    import argcomplete
+except ImportError:
+    raise ImportError('argcomplete module not found, run python3 -m pip install -r requirements.txt ')
+from argcomplete.completers import ChoicesCompleter
 
+from enum import Enum
+import os
+import consts.resource_paths
+
+def get_immediate_subdirectories(a_dir):
+    return [name for name in os.listdir(a_dir)
+            if os.path.isdir(os.path.join(a_dir, name))]
+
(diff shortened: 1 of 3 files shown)


Luxonis-Brandon:
Wow this will be super sweet to have!  Thanks for thinking of this and making it happen.  I'll try it out once I get out from under the bit of a rock I'm under!

depthai_helpers/cli_utils.py
```diff
@@ -1,8 +1,39 @@
...
+
 from enum import Enum
+import os
+import consts.resource_paths
+
+def get_immediate_subdirectories(a_dir):
+    return [name for name in os.listdir(a_dir)
+            if os.path.isdir(os.path.join(a_dir, name))]
+
+stream_choices = ("metaout", "previewout", "jpegout", "left", "right", "depth_sipp", "disparity", "depth_color_h", "meta_d2h", "object_tracker")
```
VanDavv, line 18: `video` stream should be listed here also?

VanDavv:
Super nice feature and very easy to integrate, kudos @szabi-luxonis!

depthai_helpers/cli_utils.py
VanDavv: Also, since #148 was merged, the stream names also changed

PR #171, opened 2020-08-02
Check if cameras are connected or not.

iamsg95:
Check if cameras are connected or not. If not, print warning message to let user know that the camera might be disconnected or broken.
Also if NN is running only on left/right camera the RGB pipeline is disabled, to save resources.


Luxonis-Brandon:
Thanks.  Testing now with BW1098FFC.  This will be great to have for debugging incorrectly-connected cameras on the FFC variants or FFCs that got bumped loose in use/etc.

Luxonis-Brandon:
This works great for me.  I unplugged the `CAM_R` on the BW1098FFC and got this:
image

And here is w/ the grayscale plugged back in and the RGB disconnected:
image

Luxonis-Brandon:
OK only one thing I noticed @iamsg95 :
```./depthai.py -cnn face-detection-retail-0004 -cnn2 landmarks-regression-retail-0009 -cam left_right -dd -sh 12 -cmx 12 -nce 2 -monor 400 -monof 30```

This results in the RGB error even though I think it should still run OK w/out RGB connected.

image


Thoughts?

iamsg95:
Done.
If camera is not connected we will fall back to a working mode (if not all cameras are disconnected).

Luxonis-Brandon:
image
Looks good to me - thanks!

image

image

PR #177, opened 2020-08-13
person segmentation using deeplabv3 

rt-luxonis:
deeplabv3 model with mobilenetv2 backbone trained for person semantic segmentation
run using depthai.py -cnn deeplabv3p_person
For syncing the camera frame rate with nn detection, add the parameters: -sync -rgbf 12 , where 12 should match the nn frame rate, or be just below it.


Luxonis-Brandon:
For tracking purposes, this is implemented in: https://github.com/luxonis/depthai/pull/202

PR #178, opened 2020-08-14
Enhance stereo depth

alex-luxonis:
Add options to configure:
- the kernel size of the median filter that is applied over the disparity map: 3x3, 5x5, 7x7 (default), or disabled.
- enabling the `Left-Right check` feature (runs disparity on both directions and combines them).
```
  -med {0,3,5,7}, --stereo_median_size {0,3,5,7}
                        Disparity / depth median filter kernel size (N x N) .
                        0 = filtering disabled. Default: 7
  -lrc, --stereo_lr_check
                        Enable stereo 'Left-Right check' feature.
```
Currently, the disparity map works well (`disparity`, `disparity_color` streams), but there is a problem with `depth_raw` and the `-lrc` option (WIP).
(It's also better to use a disparity stream to visualize the map quality).

Currently, with `-lrc`, the median filtering (if not disabled) is applied on each direction (L-R and R-L), before combining. We may need to move it after the combine step.

Example usage: `./depthai.py -s disparity_color -med 5 -lrc`


alex-luxonis:
Merged in the branch `metadata_disparity_depth`, with these changes:
- added frame metadata for `disparity`, `disparity_color`, `depth_raw` streams too. Previously it was present only for `previewout`, `left`, `right`, `metaout` (NN packets).
- fixed streaming out `depth_raw` with the full resolution 1280×800 (`-monor 800` option)

VanDavv:
@alex-luxonis can we close this PR?

alex-luxonis:
Will rebase the work here on develop, then close it. The flicker should only happen with `-lrc` which is a new option (and I'll fix it).
About the `metadata_disparity_depth` branch, that one was merged into this. So now `metadata_disparity_depth` is obsolete.

saching13 (approved):

Luxonis-Brandon:
I think we can close this now.  So closing now but we can always reopen.

PR #179, opened 2020-08-17
Develop merge config manager

jonngai:
I ending up moving most of the argument parsing logic in the demo python script to their own classes to make reuse for ROS2 a bit cleaner.


depthai_helpers/arg_manager.py
```diff
@@ -0,0 +1,189 @@
...
+    option_args = len(option_list)
+    if option_args not in [1, 2]:
+        msg_string = "{0} format is invalid. See --help".format(option)
+        cli_print(msg_string, PrintColors.WARNING)
+        raise ValueError(msg_string)
+
+    deprecated_choices = ("depth_sipp", "depth_color_h")
+    transition_map = {"depth_sipp" : "disparity_color", "depth_color_h" : "disparity_color"}
+    stream_name = option_list[0]
+    if stream_name in deprecated_choices:
```
VanDavv, line 32: Can be simplified
```suggestion
    if stream_name in transition_map:
```

depthai_helpers/config_manager.py
```diff
@@ -0,0 +1,342 @@
...
+from depthai_helpers.model_downloader import download_model
+from depthai_helpers.cli_utils import cli_print, PrintColors
+
+class DepthConfigManager:
+    labels = ""
+
+    def __init__(self, args):
+        self.args = args
+        self.stream_list = args['streams']
+        self.compile_model = self.args['shaves'] is not None and self.args['cmx_slices'] is not None and self.args['NN_engines']
```
VanDavv, line 17: should be `or` instead of `and` (as if I change just `NN_engines` the model should be also compiled

depthai_helpers/config_manager.py
```diff
@@ -0,0 +1,342 @@
...
+        self.calc_dist_to_bb = not self.args['disable_depth']
+
+        # Prepare handler methods (decode_nn, show_nn) for the network we want to run.
+        self.decode_nn, self.show_nn, self.decode_nn_json = self.importAndGetCallbacksForNN()
```
VanDavv, line 21: maybe it would be cleaner/easier to use `@property` annotation (for `self.jsonConfig` also)?

VanDavv (commented):
I doubt there are 166 commits to be merged. Best would be to both rebase `develop` and your branch with the latest commit so that the diff may be clearer

VanDavv:
I've also noticed a bug @jonngai 

```
$ python depthai_demo.py -s previewout metaout left right disparity disparity_color -cmx 14 -sh 14 -nce 2 -brd bw1098obc
Using depthai module from:  /home/vandavv/dev/luxonis/depthai-python/venw/lib/python3.8/site-packages/depthai.cpython-38-x86_64-linux-gnu.so
Using Custom Calibration File: depthai.calib
  File "depthai_demo.py", line 307, in <module>
    dai.startLoop()
  File "depthai_demo.py", line 51, in startLoop
    configMan = DepthConfigManager(args)
  File "/home/vandavv/dev/luxonis/depthai/depthai_helpers/config_manager.py", line 22, in __init__
    self.jsonConfig = self.generateJsonConfig()
  File "/home/vandavv/dev/luxonis/depthai/depthai_helpers/config_manager.py", line 115, in generateJsonConfig
    blobMan = BlobManager(self.args, self.compile_model, self.calc_dist_to_bb)
  File "/home/vandavv/dev/luxonis/depthai/depthai_helpers/config_manager.py", line 273, in __init__
    self.blob_file, self.default_blob = self.compileBlob(blob_file)
NameError: name 'blob_file' is not defined
```

VanDavv:
Another one

```
$ python depthai_demo.py -s previewout metaout left right disparity disparity_color -brd bw1098obc
Using depthai module from:  /home/vandavv/dev/luxonis/depthai-python/venw/lib/python3.8/site-packages/depthai.cpython-38-x86_64-linux-gnu.so
Using Custom Calibration File: depthai.calib
  File "depthai_demo.py", line 307, in <module>
    dai.startLoop()
  File "depthai_demo.py", line 51, in startLoop
    configMan = DepthConfigManager(args)
  File "/home/vandavv/dev/luxonis/depthai/depthai_helpers/config_manager.py", line 22, in __init__
    self.jsonConfig = self.generateJsonConfig()
  File "/home/vandavv/dev/luxonis/depthai/depthai_helpers/config_manager.py", line 205, in generateJsonConfig
    config = self.postProcessJsonConfig(config)
  File "/home/vandavv/dev/luxonis/depthai/depthai_helpers/config_manager.py", line 221, in postProcessJsonConfig
    utils.merge(board_config, config)
NameError: name 'utils' is not defined
```

VanDavv:
I would also merge `tests` and `testsFunctional` directory (`test_cli_utils.py` can be also a part of functional tests)

VanDavv:
Path resolution for `runFunctionalTests.py` is incorrect, it should be relative to the file itself, not to the runtime context.
It works when inside `testsFunctional` directory, but when I'm in the root `depthai` folder, this is the output I get

```
$ python testsFunctional/runFunctionalTests.py 
python3: can't open file 'test_meta_d2h.py': [Errno 2] No such file or directory
----------------------------------------------------------------
meta_d2h Failed...
----------------------------------------------------------------
python3: can't open file 'test_object_tracker.py': [Errno 2] No such file or directory
----------------------------------------------------------------
object_tracker Failed...
----------------------------------------------------------------
----------------------------------------------------------------
Running previewout
----------------------------------------------------------------
python3: can't open file 'test_video.py': [Errno 2] No such file or directory
----------------------------------------------------------------
previewout Failed...
----------------------------------------------------------------
----------------------------------------------------------------
Running left
----------------------------------------------------------------
python3: can't open file 'test_video.py': [Errno 2] No such file or directory
----------------------------------------------------------------
left Failed...
----------------------------------------------------------------
----------------------------------------------------------------
Running right
----------------------------------------------------------------
python3: can't open file 'test_video.py': [Errno 2] No such file or directory
----------------------------------------------------------------
right Failed...
----------------------------------------------------------------
----------------------------------------------------------------
Running depth_raw
----------------------------------------------------------------
python3: can't open file 'test_video.py': [Errno 2] No such file or directory
----------------------------------------------------------------
depth_raw Failed...
----------------------------------------------------------------
----------------------------------------------------------------
Running disparity
----------------------------------------------------------------
python3: can't open file 'test_video.py': [Errno 2] No such file or directory
----------------------------------------------------------------
disparity Failed...
----------------------------------------------------------------
----------------------------------------------------------------
Running disparity_color
----------------------------------------------------------------
python3: can't open file 'test_video.py': [Errno 2] No such file or directory
----------------------------------------------------------------
disparity_color Failed...
----------------------------------------------------------------
  File "testsFunctional/runFunctionalTests.py", line 69, in <module>
    cnnlist = os.listdir(cnndir)
FileNotFoundError: [Errno 2] No such file or directory: '../resources/nn/'
```

VanDavv:
I've confirmed that tests work well with BW1098OBC.
However testing out on BW1093 (MegaAI), it started to loop on the left stream test.
So two suggestions here:
- After 2-3 failed attempts, the test should fail and continue next ones
- Some tests should only be run with the correct hardware (so left/right/disparity/disparity_color streams should not be tested if run against BW1093)

depthai_helpers/arg_manager.py
jonngai: I'm not sure if I understand this simplification. It seems like switching deprecated_choices and transition_map in this conditional would cause totally different behavior.

depthai_helpers/config_manager.py
jonngai: That makes sense, I suppose. I just moved the existing logic which requires all three to be passed in order to have that compile variable set. I can change it to do that though. Just make it default to
            shave_nr = 7
            cmx_slices = 7
            NCE_nr = 1
if nothing is passed for one of the three?

depthai_helpers/config_manager.py
jonngai: Hmm, yeah, I guess returning that many variables isn't pretty. l can change these to set the class variables in the methods instead.

jonngai:
Thanks a ton for taking a look at this @VanDavv. I've fixed the two bugs and the issue with the script run path.

>I would also merge tests and testsFunctional directory (test_cli_utils.py can be also a part of functional tests)
test_cli_utils.py was actually a unit test. I felt like keeping them separate because of that but if you want to veto that choice I can move them to the same directory.

>However testing out on BW1093 (MegaAI), it started to loop on the left stream test.
Interesting. I don't have a BW1093 but I'm assuming the loop it's getting stuck in is not the main loop where it retrieves packets? Is there an easy way to reproduce this without a BW1093?

Luxonis-Brandon:
Yes, if you unplug the grayscale cameras from your board it should have the same behavior.  Also we'll get you a unit.

jonngai:
Is there any way to detect what model of device we have hooked up from the scripts? It would actually be cooler if we just told the user that the left, right, disparity, etc streams weren't available for the device that they have in the depthai_demo.py. Looks like we give the same usb vendor id for both BW1093 and BW1098OBC.

VanDavv:
> I'm assuming the loop it's getting stuck in is not the main loop where it retrieves packets? Is there an easy way to reproduce this without a BW1093?
Yes, the source code tries to create a pipeline with the stream `left`, whereas the BW1093 does not this stream and crashes - which causes the test to start over instead of failing
> Is there any way to detect what model of device we have hooked up from the scripts
I know there is EEProm data we flash on our devices, which also contain the board name - we can use it. I believe that @iamsg-luxonis also did the checks for the left/right cameras if they are connected [here](https://github.com/luxonis/depthai/pull/171), so maybe we can also expose this information through python bindings

depthai_helpers/arg_manager.py
VanDavv: So in this case we want to check if one of the depreciated streams was passed by the user. We have transition map with keys being the old stream names and values being the new stream names

When we do `if key in dict` we check if the key is present in the dict's keys. So in this case, as you have listed all depreciated stream names as dict keys, you can go ahead and do `if stream in transition_map`.

`depreciated_choices`  is the same as `transition_map.keys()`, so `if stream_name in depreciated_choices` is equal to `if stream_name in transition_map.keys()`.
And since comparison `if stream_name in transition_map.keys()` is the same as `if stream_name in transition_map`, we can remove the `depreciated_choices` array and stick with just `transition_map`

depthai_helpers/config_manager.py
VanDavv: yes, good call!

depthai_helpers/config_manager.py
VanDavv: For the `self.jsonConfig`, you could do

```python
@property
def jsonConfig(self):
    # body of postProcessJsonConfig method
```

and then when you access `self.jsonConfig`, you treat it as a regular variable, but it will actually run the function to resolve it's value.
But for the `decode_nn` and others - it's fine for me to leave it as it is, I just didn't like that we assign to `self` both here and inside `self.importAndGetCallbacksForNN()`

depthai_helpers/arg_manager.py
jonngai: Oh, gotcha, didn't notice they contained the same keys. Yeah, I'll get rid deprecated_choices.

Luxonis-Brandon:
Yes, agreed this will be very useful.  So the BW1093 is the only model without an eeprom.  All the others can be identified by the eeprom data, and we can safely assume that if there is no eeprom that it is the BW1093.  

I agree it would be great to be able to pull this information from Python bindings.  I'm not sure if it exists yet.

jonngai:
Yes, as it stands right now I think additional changes in depthai-core are required to prevent this infinite retry. I agree that it would be good to be able to grab the board ids from the python API and think there's value to that beyond just getting these tests to pass. 

If possible, I'd like to check this particular set of changes in so other people can try it out and find more bugs. In the meantime, I can look into adding/finding the bindings necessary to have the script and C++ API tell the user certain cameras aren't available rather than trying to connect to a camera stream to figure out it's missing from a board.

Does that sound reasonable?

Luxonis-Brandon:
Yes, sounds great to me @jonngai .

PR #188, opened 2020-09-04
Create bw1092.json to allow calibration and eeprom settings

Luxonis-Brian:
Will `-brd bw1092` pick this up? Or are there more changes required?


saching13 (commented):
I don't think any changes is needed. It should be able to pick it up. I will run a test and cross check once just to double check

saching13:
Yes it works. it is capable of detecting and reading the json but I think it's better to have the filename in caps since others are also in caps.

Luxonis-Brian:
Good idea on the consistency. I made the change. I'll go ahead and merge.

PR #190, opened 2020-09-10
Enhance stereo depth develop merge

iamsg-luxonis:
Pull request #178 merged into develop.
Run:
`python3 -m pip install -r requirements.txt`


depthai_demo.py
```diff
@@ -216,6 +237,14 @@ def startLoop(self):
                             cv2.putText(frame, packet.stream_name, (25, 25), cv2.FONT_HERSHEY_SIMPLEX, 1.0, (0, 0, 255))
                             cv2.putText(frame, "fps: " + str(frame_count_prev[window_name]), (25, 50), cv2.FONT_HERSHEY_SIMPLEX, 1.0, (0, 0, 255))
                         else: # uint16
+                            if args['pointcloud'] and "depth_raw" in stream_names and "rectified_right" in stream_names and right_rectified is not None:
+                                if pcl_not_set:
+                                    pcl_converter = PointCloudVisualizer(depthai.get_right_intrinsic(), 1280, 720)
```
jonngai: I think we may be missing an import for this?

depthai_helpers/config_manager.py
```diff
@@ -30,6 +30,7 @@ def getUsb2Mode(self):
             usb2_mode = True
         else:
             usb2_mode = False
+        return usb2_mode
```
jonngai, line 33: Whoops, thanks for fixing this.

depthai_demo.py
jonngai: I also get this error:
AttributeError: module 'depthai' has no attribute 'get_right_intrinsic'

I did a python3 -m pip install -r requirements.txt but maybe a version mismatch?

saching13: Yes. there is a version mismatch. Need to fix that.

saching13: Yes. I added that already

iamsg-luxonis: Fixed it.
When updating `depthai-python`  `requirements.txt` should be updated too.
See commit : 
6aee211

iamsg-luxonis: There is issue after update though @saching13 
```
legacy, get_Translation() is not available in version 3
 recalibrate and load the new calibration to the device. 
```

iamsg-luxonis: Nevermind, I fixed it.

saching13 (approved):
Looks great. Tested. depth_raw, disparity, pcl on different resolution. mirror_frame too along with default AI model. on both version 4 and 5 of eeprom.

PR #192, opened 2020-09-16
Rename depth_raw to depth

iamsg-luxonis:
Stream `depth_raw` has been renamed to `depth` to match naming convention with other streams.


Luxonis-Brandon (approved):
LGTM

PR #193, opened 2020-09-18
Two stage NN + calibration fix

szabi-luxonis:
Two stage NN was broken due to a small error in refactorization done for arg/config manager.

```
./depthai_demo.py -cnn face-detection-retail-0004 -cnn2 landmarks-regression-retail-0009 -cam left_right -sh 12 -cmx 12 -nce 2 
```
was broken.

The problem appeared only when the model was auto-compiled.

Calibration was also fixed when EEPROM was empty.

### DIFF
--- a/depthai_helpers/config_manager.py
+++ b/depthai_helpers/config_manager.py
@@ -291,9 +291,9 @@ class BlobManager:
         # compile modules
         self.default_blob=True
         if compile_model:
-            self.blob_file, self.default_blob = self.compileBlob(self.blob_file)
+            self.blob_file, self.default_blob = self.compileBlob(self.args['cnn_model'])
             if self.args['cnn_model2']:
-                self.blob_file2, self.default_blob = self.compileBlob(self.blob_file2)
+                self.blob_file2, self.default_blob = self.compileBlob(self.args['cnn_model2'], False)
 
     def getLabels(self):
         # try and load labels
@@ -331,7 +331,9 @@ class BlobManager:
 
         return blobFile, blobFileConfig
 
-    def compileBlob(self, blob_file):
+    def compileBlob(self, nn_model, isFirstNN=True):
+        blob_file, _ = self.getBlobFiles(nn_model, isFirstNN)
+
         default_blob=False
         shave_nr = 7 if self.args['shaves'] is None else self.args['shaves']
         cmx_slices = 7 if self.args['cmx_slices'] is None else self.args['cmx_slices']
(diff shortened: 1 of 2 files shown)


jonngai:
Thanks for finding this. Looks good at a glance, but I'll give it a test in the morning as well.

depthai_helpers/config_manager.py
```diff
@@ -291,9 +291,9 @@ def __init__(self, args, compile_model, calc_dist_to_bb):
         # compile modules
         self.default_blob=True
         if compile_model:
-            self.blob_file, self.default_blob = self.compileBlob(self.blob_file)
+            self.blob_file, self.default_blob = self.compileBlob(self.args['cnn_model'])
```
jonngai, line 294: Does getBlobFiles serve a purpose anymore?

szabi-luxonis: I was focusing more on the fix, but yes, it can be removed.

jonngai: That's a valid point. Feel free to check it in as soon as you think it's ready then and I can do any extra cleanup and testing in a new PR.

Luxonis-Brandon:
This update worked for me.  But it did have to do the watchdog reset 2 times before it eventually worked.  I did not do any installation of requirements.txt of any of that.  Would that have been the problem?
image

Luxonis-Brandon:
Update: Now that the model is compiled for this number of SHAVES (`-sh 12 -cmx 12 -nce 2 `), the watchdog reset is not occurring.  Tried 4 times and did not see the watchdog a single time.

Luxonis-Brandon:
Trying again now w/ 8 SHAVES and 8 CMX slices to force a model recompile...  And that also does not have the watchdog trigger.  Not sure how/why it happened initially and not again.  This is being tested on a BW1092 which has USB2 only.

szabi-luxonis:
The issue mentioned is still present on develop. Didn't track back yet when it was introduced. @Luxonis-Brandon

Luxonis-Brandon:
Thanks.  Good to know.  Curious what the root is here.  Have we confirmed that this does not occur on `main`?

Luxonis-Brandon:
(I'm going to try repeating several times on main to see if I can get it to occur.)

Luxonis-Brandon:
image

Going back to main - one thing I just noticed is it seems like we switched from `depthai-demo.py` (in `main`) to `depthai_demo.py` (in `develop`).

Not sure what the reasoning is here.  But would be good to keep one and not switch again.  All our documentation is using `depthai_demo.py` actually.  

https://docs.luxonis.com/tutorials/openvino_model_zoo_pretrained_model/

Luxonis-Brandon:
OK, I can confirm that this issue does not occur on `main`.

jonngai:
@Luxonis-Brandon Sorry, I renamed that file. You can't import from files with a "-" in them. It's probably good practice to just avoid using hyphens in python file names.

depthai_helpers/config_manager.py
szabi-luxonis: Actually it has a purpose, when `compile_model` is false it provides the default config.

PR #196, opened 2020-09-18
open3d install and point cloud conversion for raspberry pi disabled.

saching13:
Since raspberry pi doesn't have a prebuilt open3d library. Conversion to point cloud and visualization is disabled for now.

### DIFF
--- a/depthai_demo.py
+++ b/depthai_demo.py
@@ -58,7 +58,7 @@ class DepthAI:
 
         configMan = DepthConfigManager(args)
         if is_rpi and args['pointcloud']:
-            warnings.warn("Open3D is not available on raspberry pi so point cloud is disabled", ImportWarning)
+            print("Open3D is not available on raspberry pi so point cloud is disabled.")
         # these are largely for debug and dev.
         cmd_file, debug_mode = configMan.getCommandFile()
         usb2_mode = configMan.getUsb2Mode()
--- a/depthai_demo.py
+++ b/depthai_demo.py
@@ -10,8 +10,9 @@ from datetime import datetime
 import cv2
 import numpy as np
 import sys
-
+import warnings
 import depthai
+
 print('Using depthai module from: ', depthai.__file__)
 
 import consts.resource_paths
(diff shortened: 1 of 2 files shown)


depthai_demo.py
```diff
@@ -50,7 +56,8 @@ def startLoop(self):
         args = vars(cliArgs.parse_args())
 
         configMan = DepthConfigManager(args)
-
+        if is_rpi and args['pointcloud']:
+            print("Open3D is not available on raspberry pi so point cloud is disabled.")
```
szabi-luxonis, line 60: Use:
```
raise NotImplementedError("Point cloud visualization is currently not supported on RPI")
```

depthai_demo.py
```diff
@@ -238,7 +245,7 @@ def print_packet_info(packet):
                             cv2.putText(frame, packet.stream_name, (25, 25), cv2.FONT_HERSHEY_SIMPLEX, 1.0, (0, 0, 255))
                             cv2.putText(frame, "fps: " + str(frame_count_prev[window_name]), (25, 50), cv2.FONT_HERSHEY_SIMPLEX, 1.0, (0, 0, 255))
                         else: # uint16
-                            if args['pointcloud'] and "depth" in stream_names and "rectified_right" in stream_names and right_rectified is not None:
+                            if not is_rpi and args['pointcloud'] and "depth" in stream_names and "rectified_right" in stream_names and right_rectified is not None:
```
szabi-luxonis, line 248: remove, keep the old. We shouldn't reach here. IF user requests pcl and it's not available we should inform him.

szabi-luxonis (changes requested):

depthai_demo.py
saching13: Okay.

depthai_demo.py
saching13: I was just returning a print so I changed it here so that it will not break. but sure. will remove it here since will reaise an error above.

szabi-luxonis:
LGTM

szabi-luxonis (approved):

PR #209, opened 2020-09-28
Add ISP 3A camera controls

alex-luxonis:
Please check the source code changes in https://github.com/luxonis/depthai-python/pull/52


depthai_demo.py
```diff
@@ -337,6 +337,19 @@ def print_packet_info(packet):
                 self.device.request_af_mode(depthai.AutofocusMode.AF_MODE_AUTO)
             elif key == ord('2'):
                 self.device.request_af_mode(depthai.AutofocusMode.AF_MODE_CONTINUOUS_VIDEO)
+            # 5,6,7,8,9,0 : temporary, TODO delete
+            elif key == ord('5'):
```
szabi-luxonis, line 341: create a lookup table (dictionary) for all these ISP commands, so it's easier to add/remove and the code is cleaner.
In `DepthAI` class:
```
    def keypress_handler(self, cmd):

        keypress_handler_lut = {
            ord('f'): lambda: self.device.request_af_trigger()
       . 
       . 
       .
        }
        if cmd in keypress_handler_lut:
            keypress_handler_lut[cmd]()
        return
```

instead the elif's : 
`self.keypress_handler(key)`

szabi-luxonis (changes requested):

szabi-luxonis (approved):

Luxonis-Brandon:
Sweet, this will be nice to get into develop.  Thanks, all!

PR #221, opened 2020-10-07
Option to run NN on stereo rectified streams

alex-luxonis:
Source code PR: https://github.com/luxonis/depthai-python/pull/53

New options for `-cam / --cnn_camera`: `rectified_left`, `rectified_right`, `rectified_left_right` (both L and R rectified).
Note: by default the rectified streams are mirrored, so depending on usecase this option may be needed: `-mirror_rectified false`


szabi-luxonis (approved):
LGTM

PR #224, opened 2020-10-08
Develop merge - release 0.3.0.0

VanDavv:
Awaiting https://github.com/luxonis/depthai/pull/242


Luxonis-Brandon:
Thanks.

So if #244 was already present in `main` (is it @Luxonis-Brian ?) I think we should merge before that.  And then we can hot-fix `main` based on the fix for #244 .  Thoughts?

Luxonis-Brian:
checking @Luxonis-Brandon....

szabi-luxonis:
> Thanks.
> 
> So if #244 was already present in `main` (is it @Luxonis-Brian ?) I think we should merge before that. And then we can hot-fix `main` based on the fix for #244 . Thoughts?

@alex-luxonis already fixed it, and opened a PR for it (waiting for CI to finish)

Luxonis-Brandon:
Ok perfect thanks!

alex-luxonis:
Yes, it's good to have #244  fixed, as the issue was not present in main.

PR #225, opened 2020-10-09
Check if installed depthai version matches required

szabi-luxonis:
Sometimes people forget to reinstall requirements after switching branches.
This PR adds checks to verify if installed depthai version matches the required one in `requirements.txt`.


depthai_demo.py
```diff
@@ -13,6 +13,32 @@
...
 print('Using depthai module from: ', depthai.__file__)
+print('Depthai version installed: ', depthai.__version__)
+
+def get_version_from_requirements():
+    with Path('requirements.txt').open() as f:
+        datafile = f.readlines()
+    for line in datafile:
+        if 'depthai' in line:
+            #not commented out
+            if(line.split()[0] != '#'):
```
VanDavv, line 24: it won't work if comment is done this way
```
#requests
```
I'd make
```suggestion
            if not line.startswith('#'):
```

VanDavv (changes requested):

depthai_helpers/version_check.py
```diff
@@ -0,0 +1,29 @@
...
+
+def check_depthai_version():
+    version_required = get_version_from_requirements()
+    if version_required is not None:
+        print('Depthai version required:  ', version_required)
+        if version_required != depthai.__version__:
+            raise ValueError(f"\033[1;5;31mVersion mismatch \033[0m\033[91m between installed depthai lib and the required one by the script.\033[0m \n\
+                Required:  {version_required}\n\
+                Installed: {depthai.__version__}\n\
+                \033[91mRun: ./install_requirements.sh \033[0m")
```
alex-luxonis, line 29: How should we handle this check (in a generic way) when running with a local/development version?
I'm building in `depthai-python/build` with `cmake ..` and `make -j`. 
And when running depthai_demo.py, I set an environment variable `export PYTHONPATH=...../depthai-python/build` (that I unset when testing with a prebuilt wheel: `unset PYTHONPATH`).

It fails currently with:
```
                Required:  0.2.0.1+406f48b91e9bd01824c748a0028e07595b5557dd
                Installed: 0.2.0.1
```

Maybe appending an extra suffix like `-dev` to the version built locally, when not running in a CI environment?

themarpe: That could work, and you can then just configure `depthai-python` with this special hash (or just dev): 
```cmake .. -DDEPTHAI_PYTHON_COMMIT_HASH=dev```
We could also make that default I guess (if env CI not set)
And on the PYTHONPATH, you may also do the following:
```PYTHONPATH="path/to/build" python3 depthai_demo.py```
To not need to set and unset the variable

PR #202, opened 2020-09-24
Openvino tensor representation develop merge

szabi-luxonis:
Changelog:

- Added support for YOLO networks + depth calculation on bounding boxes returned by yolo, on device side
- Added openpose, openpose2, deeplabv3_person networks
- Added support for 2 output formats : `raw` and `detection` 

"raw" is default, meaning that no computation is done on device (not even depth bounding box calculation). When there is no NN json provided by user, output format is automatically set to raw.
"detection" : for detection networks like YOLO, mobilenet. Decoding is done on device, and depth calculation is done on the detected bounding box, if requested.

See example : `resources/nn/mobilenet-ssd/mobilenet-ssd.json `
```
    "NN_config":
    {
        "output_format" : "detection",
        "NN_family" : "mobilenet",
        "confidence_threshold" : 0.5
    },
```
NN_config describes user config, specific to NN_family. In case of mobilenet only the threshold is configurable.
In case of tiny-yolo-v3 the networks requires more network specific metadata : `NN_specific_metadata`: 
```
    "NN_config":
    {
        "output_format" : "detection",
        "NN_family" : "YOLO",
        "NN_specific_metadata" :
        { 
            "classes" : 80,
            "coordinates" : 4,
            "anchors" : [10,14, 23,27, 37,58, 81,82, 135,169, 344,319],
            "anchor_masks" : 
            {
                "side26" : [1,2,3],
                "side13" : [3,4,5]
            },
            "iou_threshold" : 0.5,
            "confidence_threshold" : 0.5
        }
    },
```

- Added support for output and input tensor info : shape, offset, name.

The requested tensor now it's returned as a shaped numpy array, based on the network description.
For the new functions added to the API see:
https://github.com/luxonis/depthai-python/pull/51

PR #231, opened 2020-10-12
Remove old tensor representation from jsons

szabi-luxonis:
Remove old tensor representation from NN configs that I missed to remove in #202 .

Reported here:
https://luxonis-community.slack.com/archives/CT79Z6GP5/p1602491428119000


Luxonis-Brandon:
Thanks.  I had meant to bring this up actually but had forgotten.

PR #241, opened 2020-10-15
Luxonis model zoo

szabi-luxonis:
Forked from openvino model downloader.

Models are defined in `.yml` files, and the IR (openvino representation) files are downloaded on the first run, then compiled using `luxonis.com:8080`. 
There is a limitation of model size currently in our cloud compiler therefore Yolo-v3 currently doesn't compile, I talked to @VanDavv and we are going to raise the limit.

This downloader can be further improved, so we move everything to `luxonis.com:8080` providing only the `.yml` file.
In the `yml` file we can specify the conversion parameters  from tf,caffemodel,pytorch to IR model, so we won't be limited to openvino representation. Useful when switching openvino versions, we don't have to update the IR models (xml + bin).


szabi-luxonis:
I just noticed that I accidentally pushed a few extra files. Will cleanup the PR, then reopen it.

PR #250, opened 2020-10-21
Custom firmware option

alex-luxonis:
This is mostly intended for quick experiments that modify only the device firmware, and it helps to skip the rebuild of the python wheel.

The `FIRMWARE` commit hash value will be provided by Luxonis (to ensure it is compatible with the host app), and can be hardcoded in https://github.com/luxonis/depthai/blob/85db9ff/depthai_helpers/config_manager.py#L17 (on a given branch)
or inputted by this command line option:
```
  -fw FIRMWARE, --firmware FIRMWARE
                        Commit hash for custom FW, downloaded from Artifactory
```

The firmware specified is downloaded from [Luxonis Artifactory - device side](https://artifacts.luxonis.com/artifactory/luxonis-myriad-snapshot-local/depthai-device-side/) and cached locally inside the repo, in `.fw_cache` directory. Both normal (USB3-enabled) and USB2 variants of the firmware are supported.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,6 +31,7 @@ MANIFEST
 *.idea
 # DepthAI-Specific
 dataset/
+fw_cache/
 depthai.calib
 mesh_left.calib
 mesh_right.calib
--- a/.gitignore
+++ b/.gitignore
@@ -31,7 +31,7 @@ MANIFEST
 *.idea
 # DepthAI-Specific
 dataset/
-fw_cache/
+.fw_cache/
 depthai.calib
 mesh_left.calib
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -31,6 +31,7 @@ MANIFEST
 *.idea
 # DepthAI-Specific
 dataset/
+fw_cache/
```
szabi-luxonis, line 34: should we make cache folder hidden ? `.fw_cache`

szabi-luxonis (approved):
LGTM

PR #252, opened 2020-10-22
fixed missing blob error in calibrate.py

saching13:
With model zoo being fetched using downloader `calibrate.py` was not working since blob was not found. added a small fix to download `mobilenet-ssd` if not found during the calibration routine.

### DIFF
--- a/calibrate.py
+++ b/calibrate.py
@@ -15,7 +15,7 @@ from pathlib import Path
 import shutil
 import consts.resource_paths
 import json
-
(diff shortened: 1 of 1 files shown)


szabi-luxonis (approved):
LGTM.
Thanks for the fix, I totally missed testing the calibration script on a clean branch.

saching13:
No worries. I saw the error while updating the other branches and fixed it. :)

Luxonis-Brandon:
Yes thanks for finding and fixing @saching13 !

PR #256, opened 2020-10-27
Improve the helper scripts

VanDavv:
Our python scripts in the repo served the same purpose as bash scripts, but with much more complexity.

This PR replaces python helper scripts with their bash equivalents - so functionality stays the same and syntax is much simpler


szabi-luxonis:
Shell scripts are not cross platform, this will actually not work on windows.
Also, what happens if bash is not installed on the system? Shebang line defines `#!/bin/bash`

I recently converted the model downloader into pure python for this reason (same for all helper scripts, it would have been easier to write them in shell though)

VanDavv:
Oh, right, that's why it was so strange for me, and thought it's just some leftover.
Closing

VanDavv:
I created a separate PR just with an updated system log script - https://github.com/luxonis/depthai/pull/257

PR #257, opened 2020-10-27
Update log_system_information.sh

VanDavv:
PR #256, despite changing shell scripts, also introduced a broader system gathering script.

This PR contains only this updated script (as #256 was closed and shell migration was not good in our case)


szabi-luxonis:
It would be better to replace `log_system_information.sh` with a python script, for the same reasons mentioned in #256 .
Currently we have 3 shell scripts left over:
```
./install_requirements.sh
./package/create_package.sh
./log_system_information.sh
```
`package/create_package.sh` can be deleted.
`log_system_information.sh` should include as much things as possible.
It's actually  very simple to call a subprocess: 
Some examples:
https://github.com/luxonis/depthai/blob/main/depthai_profiler.py#L21
https://github.com/luxonis/depthai/blob/main/model_compiler/model_compiler.py#L72
`install_requirements.sh` should be a python script too (with python shebang and executable permissions) so it can be called as ./install_requirements.py

VanDavv:
Sure, changing to draft and will work in this further, thanks @szabi-luxonis!

PR #258, opened 2020-10-29
Fix & refactor tests

VanDavv:
This PR introduces new tests, derived from the previous ones that existed in our repo.

They all work now, code is consistent and uses the same mechanism the previous ones did (DepthAI as a class)

Also made a spacing and imports cleanup in depthai_demo.py.

You can run the tests using the command below
```
python3 run_tests.py
```

and it will run all tests defined in `tests/tests_runner.py` and print out the summary 
image

Streams tests verify if packet from defined stream_name has arrived, CNN tests verify if nnet_packets has arrived


tests/tests_runner.py
```diff
@@ -0,0 +1,68 @@
...
+import sys
+
+parser = argparse.ArgumentParser()
+parser.add_argument("-r", "--reruns", type=int, default=0)
+args = parser.parse_args()
+
+
+def print_result(name, success):
+    if success:
+        print(f"\033[1;34;40mTest {name}: \033[1;32;40mPassed!\033[0m")
```
szabi-luxonis, line 14: Add some global variables for these color codes, they are present in multiple lines, then use as ' f{'FLASHING_RED} .. '

run_tests.py
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env python3
+
+import subprocess
+import os
+from pathlib import Path
+import sys
+
+env = os.environ.copy()
+env["PYTHONPATH"] = str(Path(__file__).parent)
+result = subprocess.run([sys.executable, "./tests/tests_runner.py"], env=env)
```
szabi-luxonis, line 10: It would be better not using relative path here, or at least relative to 'run_tests.py' always.
Running as `depthai-host/run_tests.py` for example will fail.
```
file_path = Path(__file__).parent / "./tests/tests_runner.py"
result = subprocess.run([sys.executable, file_path], env=env)
```
should do it.

Would be nice to make our scripts callable from any folder.

tests/test_executor.py
```diff
@@ -0,0 +1,39 @@
+import argparse
+import threading
+import time
+
+from depthai_demo import DepthAI
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument("-s", "--streams", nargs="+")
```
szabi-luxonis, line 9: should we set them as `required=True`? If not then `default = None`?

szabi-luxonis (changes requested):

tests/test_executor.py
VanDavv: `default=None` is added by default by argparse

run_tests.py
VanDavv: Agreed, done (also changed filepath in requirements check)

tests/tests_runner.py
VanDavv: Added, also changed the way PrintColors are handled and parsed by `cli_print`

depthai_helpers/version_check.py
```diff
@@ -3,7 +3,8 @@
 
 
 def get_version_from_requirements():
-    with Path('requirements.txt').open() as f:
+    reqirements_path = Path(__file__).parent / Path('../requirements.txt')
```
szabi-luxonis, line 6: type, should be `requirements`

szabi-luxonis (changes requested):

themarpe (approved):
LGTM

Issue #259, opened 2020-10-31
Crash on "del device"


geaxgx:
In my program, I may need to call the following code several times in succession:

```
device = depthai.Device('', False)
pipeline = device.create_pipeline(...)
... # Some stuff
del device
```

On the 2nd call to 'del device', the program crashes:
terminate called after throwing an instance of 'std::system_error'
  what():  Argument invalide
Abandon (core dumped)

I am on ubuntu 18.04 and I can reproduce the problem easily by simply duplicating the hello_world.py code https://github.com/luxonis/depthai-tutorials/blob/master/1-hello-world/hello_world.py

alex-luxonis:
Hi @geaxgx , I can reproduce the crash, it's related to this code on C++ side, the variable `once` is already cleared at the second run, the watchdog thread is not recreated, and it will fail in the `join()` call.
https://github.com/luxonis/depthai-core/blob/678e612/src/device.cpp#L151...L167

I'll create a quick fix.

alex-luxonis:
@geaxgx Please check if the issue is resolved by reinstalling `depthai` from the branch `fix_recreate_device_loop`:
`python3 -m pip install -r requirements.txt`

And thanks for the issue report!

geaxgx:
Problem solved ! 
Thank you @alex-luxonis !

PR #260, opened 2020-10-31
Fix crash on 2nd device object delete

alex-luxonis:
Source code PR: https://github.com/luxonis/depthai-python/pull/64

Should fix issue https://github.com/luxonis/depthai/issues/259

Tested by replacing this line: https://github.com/luxonis/depthai/blob/ddfe91b7/depthai_demo.py#L382
```diff
 if __name__ == "__main__":
     dai = DepthAI()
-    dai.startLoop()
+    while True: dai.startLoop()

```
and pressing `q` during streaming, to have the loop restart.

Also tested that the watchdog still works fine, by manually resetting the device while streaming.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,4 +5,7 @@ requests==2.24.0
 argcomplete==1.12.0
 open3d==0.10.0.0; platform_machine != "armv7l"
 pyyaml==5.3.1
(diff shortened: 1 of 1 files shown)


szabi-luxonis (approved):
LGTM

PR #261, opened 2020-11-01
Update log system information script

VanDavv:
PR #256, despite changing shell scripts, also introduced a broader system gathering script.

This PR contains only this updated script (as #256 was closed and shell migration was not good in our case)


log_system_information.py
```diff
@@ -0,0 +1,47 @@
...
+    subprocess.check_call([sys.executable, "-m", "pip", "install", "pyusb"])
+    import usb.core
+
+
+def get_usb():
+    speeds = ["Unknown", "Low", "Full", "High", "Super"]
+    format_hex = lambda val: f"{val:#0{6}x}"
+    for dev in usb.core.find(find_all=True):
+        yield {
+            'port': dev.port_number,
```
szabi-luxonis, line 19: For visual consistency should be " "

szabi-luxonis (changes requested):
Lgtm other than minor request

szabi-luxonis (approved):
Lgtm

szabi-luxonis:
Don't forget to backport it to develop too @VanDavv

szabi-luxonis:
Change the target branch, first should be merged into develop.

VanDavv:
Tested on Windows, and we have issues with pyusb:
- by default, it throws "NoBackendError" as there is no USB backend to communicate (added commits to handle this)
- After solving it [here](https://stackoverflow.com/a/31332121), by installing libusb for Windows, there is no error, but the list remains empty
- Checking for answers [here](https://stackoverflow.com/a/5767737) it came out that we need to manually set the USB devices to be using libusb as their backend, otherwise they won't be listable.

Personally, I think we should just create an issue and think about solving this one later, as it still gathers more relevant data than it did before, even without a USB devices list we can have the relevant info there.

Thoughts @szabi-luxonis?

log_system_information.py
```diff
@@ -0,0 +1,49 @@
...
+def get_usb():
+    speeds = ["Unknown", "Low", "Full", "High", "Super"]
+    format_hex = lambda val: f"{val:#0{6}x}"
+    try:
+        for dev in usb.core.find(find_all=True):
+            yield {
+                "port": dev.port_number,
+                "vendor_id": format_hex(dev.idVendor),
+                "product_id": format_hex(dev.idProduct),
+                "speed": speeds[dev.speed]
```
alex-luxonis, line 23: Need to extend the `speeds` array. Newer libusb versions also report `LIBUSB_SUPER_SPEED_PLUS`. I have installed the latest on my system and it crashes with:
`IndexError: list index out of range`

alex-luxonis:
It may be useful to display that it created on disk the file `log_system_information.json`.

log_system_information.py
```diff
@@ -0,0 +1,49 @@
+import json
```
szabi-luxonis, line 1: add python shebang line and make it executable

szabi-luxonis (changes requested):

log_system_information.py
```diff
@@ -0,0 +1,49 @@
...
+    "platform": platform.platform(),
+    "processor": platform.processor(),
+    "python_build": ' '.join(platform.python_build()).strip(),
+    "python_compiler": platform.python_compiler(),
+    "python_implementation": platform.python_implementation(),
+    "python_version": platform.python_version(),
+    "release": platform.release(),
+    "system": platform.system(),
+    "version": platform.version(),
+    "win32_edition": platform.win32_edition(),
```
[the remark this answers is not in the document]
szabi-luxonis: Seems like it is related to python version.
On my ubuntu20.04 w/ python 3.8 it didn't crash. Neither on my mac.

FWIW it looks like it was just recently introduced:
"platform.win32_edition() "
https://docs.python.org/3/whatsnew/changelog.html#:~:text=platform.win32_edition()

log_system_information.py
szabi-luxonis: Also to be future proof should be something like :
```
"speed": speeds[dev.speed] if dev_speed < len(speeds) else dev_speed
```

szabi-luxonis:
Have you tested this with older python versions, like `python3.5`? @VanDavv

VanDavv:
@szabi-luxonis yes, python 3.6, 3.7, 3.8 and 3.9 on Linux + python 3.9 on Windows

szabi-luxonis:
> @szabi-luxonis yes, python 3.6, 3.7, 3.8 and 3.9 on Linux + python 3.9 on Windows
> Didn't test 3.5 as it reached it's EOL, but can do in a moment
> 
> UPDATE
> It can't work with 3.5 due to f-string usage, is it a blocker @szabi-luxonis ?

No, it's not.
You can go ahead and merge it. We are going to fix it for python3.5, since the fix is simple and ubuntu 16.04 (still supported until 2021) comes with python 3.5.

szabi-luxonis (approved):

Luxonis-Brandon:
Agreed.  ^

Issue #263, opened 2020-11-01
Model Downloader fails on Windows 10 and Python 3.9


VanDavv:
I tried to run `depthai_demo.py` on Windows 10 and Python 3.9.

This is the error I've got

```
$ C:\\Users\\VanDa\\AppData\\Local\\Programs\\Python\\Python39\\python.exe depthai_demo.py
No calibration file. Using Calibration Defaults.
Using depthai module from:  C:\Users\VanDa\AppData\Local\Programs\Python\Python39\lib\site-packages\depthai.cp39-win_amd64.pyd
Depthai version installed:  0.3.0.0+dev
Depthai version required:   0.3.0.0
Depthai development version found, skipping check.
Compiling model for 14 shaves, 14 cmx_slices and 1 NN_engines
model_compilation_target: cloud
Traceback (most recent call last):
  File "C:\Users\VanDa\dev\depthai\depthai_demo.py", line 370, in <module>
    dai.startLoop()
  File "C:\Users\VanDa\dev\depthai\depthai_demo.py", line 45, in startLoop
    configMan = DepthConfigManager(args)
  File "C:\Users\VanDa\dev\depthai\depthai_helpers\config_manager.py", line 26, in __init__
    self.generateJsonConfig()
  File "C:\Users\VanDa\dev\depthai\depthai_helpers\config_manager.py", line 192, in generateJsonConfig
    blobMan = BlobManager(self.args, self.calc_dist_to_bb, shave_nr, cmx_slices, NCE_nr)
  File "C:\Users\VanDa\dev\depthai\depthai_helpers\config_manager.py", line 377, in __init__
    self.blob_file = self.compileBlob(self.args['cnn_model'], self.args['model_compilation_target'])
  File "C:\Users\VanDa\dev\depthai\depthai_helpers\config_manager.py", line 431, in compileBlob
    ret = download_and_compile_NN_model(nn_model, model_zoo_folder, shave_nr_opt, cmx_slices_opt, NCE_nr, outblob_file, model_compilation_target)
  File "C:\Users\VanDa\dev\depthai\model_compiler\model_compiler.py", line 133, in download_and_compile_NN_model
    download_location = download_model(model, model_zoo_folder)
  File "C:\Users\VanDa\dev\depthai\model_compiler\model_compiler.py", line 28, in download_model
    result = subprocess.run(downloader_cmd)
  File "C:\Users\VanDa\AppData\Local\Programs\Python\Python39\lib\subprocess.py", line 501, in run
[...]

PR #264, opened 2020-11-01
Use the same interpreter for child processes as for parent

szabi-luxonis:
Should fix the issue #263  when the script is launched with other than `python3`, for example `python3.9`, where `python3` != `python3.9`


VanDavv:
Tested, works, thanks @szabi-luxonis!

PR #270, opened 2020-11-05
subprocess.run command doesn't have capture_output arg till python3.7 

saching13:
checking python version to determine whether to use  `capture_output` as arg to subprocess.run or not in model_compiler.py

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -48,3 +48,4 @@ log_system_information.json
 *.mvcmd
 *.so
 *.prof
+*csv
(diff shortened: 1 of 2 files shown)


szabi-luxonis:
Thanks !
I'm working on a different branch cleaning up/refactoring the model compiler to prepare for blobconverter integration. I noticed this too w/ python 3.6 and fixed it. FWIW to fix this is enough to remove `capture_output` from subprocess.run. so only : `subprocess.run(downloader_cmd)`.

saching13:
Ok cool. I will cancel this then.

PR #272, opened 2020-11-09
Model compiler refactor, add support for openvino 2020.1, 2020.4, 2021.1

szabi-luxonis:
Added support for openvino 2020.1, 2020.4, 2021.1 in the same script. Option: `-op --openvino_version`, `choices=["2020.1", "2020.4", "2021.1"]`

Refactored model_compiler script:

Currently only local compilation is supported, so this PR paves the road for cloud compilation implementation and testing for 2020.4 and 2021.1

It was tested only on Ubuntu 20.04 and 18.04. When cloud compiler is implemented I will do a more thorough testing.


Luxonis-Brandon (approved):
Tested with a variety of models and couldn't find a problem.  So this at least does not seem to break existing functionality so LGTM.

Luxonis-Brandon:
https://discuss.luxonis.com/d/88-model-download-timeout-download-01-org

We should see about this download issue from openvino when we implement this.

PR #276, opened 2020-11-13
Change behavior of image showing loop

kota-rr:
Suggestion on the image showing loop exiting behavior

The whole process is killed by ESC button, and there are no keys to exit the image showing loop in `show_rectified_images_two_calib`.
This is for making ESC only to exit the loop and the process will be kept on after exiting the loop.

Before
- [Space]: next image
- [ESC]: quit the process
- Other keys: no operation

Suggestion
- [ESC]: exit loop
- Other keys: next image

### DIFF
--- a/depthai_helpers/calibration_utils.py
+++ b/depthai_helpers/calibration_utils.py
@@ -482,6 +482,7 @@ class StereoCalibration(object):
 
         print("Average Epipolar Error: " + str(epi_error_sum / len(imgpoints_r)))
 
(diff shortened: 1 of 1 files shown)


szabi-luxonis (approved):
Thanks. The changes make sense to me.
LGTM from my part.
CC: @saching13

Luxonis-Brandon:
Thanks @kota-rr agreed your proposed UX is better.  Thanks for taking the time to code this up and submit the PR!

saching13 (approved):
LGTM too.

PR #278, opened 2020-11-17
Update to using latest python wheels

alex-luxonis:
... after merging https://github.com/luxonis/depthai-python/pull/56

Production support: 
 - USB info (speed, SN)
 - connected cams info
 - write EEPROM at runtime
 - fix crash with calib data zero


szabi-luxonis (approved):
LGTM

PR #281, opened 2020-12-07
Use --user flag for system-wide installation

VanDavv:
This PR adds the `--user` flag to install commands if the installation is system-wide (not from virtualenv)

Since the virtualenv itself does not allow the `--user` flag, failing with error

```
ERROR: Can not perform a '--user' install. User site-packages are not visible in this virtualenv.
```
an additional check is added to determine if the interpreter used by the user is from virtualenv or not

Tested on Linux (both system-wide and virtualenv) and Mac (both system-wide and virtualenv)

### DIFF
--- a/install_requirements.py
+++ b/install_requirements.py
@@ -3,7 +3,14 @@
 import subprocess
 import sys
 
-subprocess.check_call([sys.executable, "-m", "pip", "install", "pip", "-U"])
+# https://stackoverflow.com/a/58026969/5494277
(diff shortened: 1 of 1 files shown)


Luxonis-Brandon (approved):
Tested this on a brand new install of MacOS.  Commands used after install of MacOS

```
/bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install.sh)"
git clone https://github.com/luxonis/depthai.git
git checkout install_requirements_user
python3 install_requirements.py
```
This succeeded.  

So I think this is good to go.

PR #287, opened 2020-12-18
Increase RGB/Mono manual exposure max limit

alex-luxonis:
It was previously limited to about 33ms from device side.
To set larger exposures, the FPS need to be reduced accordingly (depthai_demo.py options `-rgbf` / `-monof`), for example 5 fps for max 200ms exposure.

Use `S` key to toggle between the cameras (`rgb` or `mono`) where the exposure control is applied. And change with:
```
            # Control:      key[dec/inc]  min..max
            # exposure time:     i   o    1..(1000*1000/fps) [us]
            # sensitivity iso:   k   l    100..1600
```
For mono cameras, a flicker artifact could appear in the frame following a change in the exposure/iso value. This needs to be debugged.

For mono cameras, currently the max exposure that can be applied:
- 0.596s at  640x400
- 1.252s at 1280x720 or 1280x800 - however with the fps set below 1.0 there's a risk the watchdog could kick in and do a reset. Workaround: add `meta_d2h` to the list of streams, for example: `-s left right meta_d2h -monof 0.5`. Also if the FPS is set to lower than the camera can output, it will be capped accordingly, for example `0.5` here results in an actual `0.79`

### DIFF
--- a/depthai_demo.py
+++ b/depthai_demo.py
@@ -138,6 +138,9 @@ class DepthAI:
             return
 
         def keypress_handler(self, key, stream_names):
+            cams = ['rgb', 'mono']
+            self.cam_idx = getattr(self, 'cam_idx', 0)  # default: 'rgb'
+            cam = cams[self.cam_idx]
             cam_c = depthai.CameraControl.CamId.RGB
             cam_l = depthai.CameraControl.CamId.LEFT
             cam_r = depthai.CameraControl.CamId.RIGHT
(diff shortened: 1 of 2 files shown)


Luxonis-Brandon (approved):
image
Tested this briefly with as dark as I could make my office (all lights off, no external lighting other than my computer monitor) and it seems to work quite well.

szabi-luxonis (approved):
LGTM

alex-luxonis:
Source code changes (just FW commit update): 
https://github.com/luxonis/depthai-python/pull/73
https://github.com/luxonis/depthai-core/pull/31

PR #296, opened 2021-01-05
added orientation check in calibration

saching13:
## Camera orientation check in calibration
- Checks if the `swap_left_and_right_cameras` is set appropriately so that left and right cameras are not interchanged.
- This scenario is most common when using the Modular camera which can be overcome by checking the distance of the pixels between two points on two images of the calibration checkerboard. 
- Product description [here](https://docs.luxonis.com/en/latest/pages/products/#bw1098ffc)


calibrate.py
```diff
@@ -340,6 +379,12 @@ def capture_images(self):
                     frame_list.append(small_frame)
 
                     if captured_left and captured_right:
+                        print("Images captured-->")
```
szabi-luxonis, line 382: These could be on the same line: `print("Images captured-->"+str(self.images_captured))`

szabi-luxonis (changes requested):
Other than minor observation LGTM

calibrate.py
```diff
@@ -265,6 +283,25 @@ def show(position, text):
...
+
+        show((60, int(height / 2 - 40)), "Calibration failed, Left and ")
+        show((60, int(height /2)), "right camera are swapped!")
+        show((60, int(height / 2 + 40)), "Fix \"swap_left_and_right_cameras\"")
+        show((60, int(height / 2 + 80)), "and start again")
+
+        # cv2.imshow("left", info_frame)
+        # cv2.imshow("right", info_frame)
+        cv2.imshow("left + right",info_frame)
+        cv2.waitKey(5000)
```
VanDavv, line 303: Maybe it would be better to use `cv2.waitKey(0)` that will just wait for a keypress without a time limit

calibrate.py
```diff
@@ -340,6 +379,12 @@ def capture_images(self):
                     frame_list.append(small_frame)
 
                     if captured_left and captured_right:
+                        print("Images captured-->")
+                        print(self.images_captured)
+                        if self.images_captured == 0:
+                            is_orientation_correct = test_camera_orientation(captured_left_frame, captured_right_frame)
```
VanDavv, line 385: can be simplified:
```
if not self.images_captured and not test_camera_orientation(...):
    self.show_failed_orientation()
```

VanDavv (changes requested):
Some suggestions, nothing serious though, LGTM

calibrate.py
VanDavv: even `print(f"Images captured --> {self.images_captured}")`

saching13: got it

saching13:
Modified the requested changed. Thanks for the suggestions.

szabi-luxonis (approved):
LGTM

Issue #200, opened 2020-09-20
On-Device 180-Degree Image Rotation (i.e. Camera Op/Down Flip)


Luxonis-Brandon:
### Start with the `why`:

In some physical installations it is advantages (or necessary) to install DepthAI and/or megaAI upside-down.  In these cases, the resultant output image needs to be 'flipped' (i.e. rotated 180 degrees).  This can be done via `cv2.flip` on the host, but doing so adds unnecessary host-burden as since megaAI/DepthAI have direct access to the image sensors settings, it is possible to reconfigure the image sensors such that there is a 'no-compute-cost' 180-degree rotation of the images coming off of the image sensors.

And such a host solution is not possible for embedded applications, where the host (which could be, for example, and ATTiny8) is not capable of doing such operations.

And even more importantly, for operations that occur on-device (such as object detection, facial-landmark detections, etc.) often the network performance degrades (or is completely non-functional) if the input image is 'upside down' of how the network was trained.

### Move to the `how`:

Implement sensor-level image flipping (i.e. 180 rotation) by reconfiguring the sensor output format and expose this capability through the DepthAI API.

### Move to the `what`:

Provide a no-compute-load capability for the images coming off the image sensors to be rotated 180-degrees to allow for 'upside-down' installation of DepthAI and megaAI.

Luxonis-Brandon:
From internal notes: 

> We could configure this directly in the camera sensor. It's straightforward for Mono (1 or 2 register sets), but will require tweaking homography matrix for StereoDepth. And for Color we need also to change the Bayer order in ISP.

And as a note, we should definitely implement this for megaAI/OAK-1 only to start (i.e. no grayscale/homography support) as the only requests we have gotten for this feature is for megaAI / OAK-1 since the color sensor is effective upside down  on that unit.

Luxonis-Brandon:
I think this is working as of today.  Checking with the team on it.

Luxonis-Brandon:
Discuss.luxonis.com comment on this for reference later: https://discuss.luxonis.com/d/71-upside-down-image-from-oak-1-camera

alex-luxonis:
Done for Gen1 on https://github.com/luxonis/depthai/pull/297

Also done for Gen2, but needs first this CI pipeline to finish: https://github.com/luxonis/depthai-python/runs/1654428842
When done, the depthai library including the fix can be installed with:
[...]

PR #297, opened 2021-01-06
Gen1: flip camera view on OAK-1

alex-luxonis:
Updated depthai lib/firmware to rotate the RGB camera view by 180° on OAK-1, and obtain a normal view with USB Type-C connector downwards.

Should fix issue https://github.com/luxonis/depthai/issues/200


Luxonis-Brandon (approved):
Works great for me!  Tried OAK-1 and BW1092, to make sure they are both the right orientation.  And they both are.  

image
image

PR #303, opened 2021-01-15
Bump opencv version in requirements

szabi-luxonis:
Bump opencv version (and matching numpy version) to the latest.
When `4.4.0.46` was released Mac OS big sur was not out.
Tested on all platforms (RPI, Linux, Windows, MacOS)


Luxonis-Brandon (approved):
LGTM

Issue #305, opened 2021-01-16
segfault during calibration of OAK-D


albertoa:
OAK-D
Ubuntu 18.04
The installation of reqs and depthai are up to date as of today.
Connected via USB 2 port

Whether I use `./calibrate.py -s 2.4 ` or `./calibrate.py -s 2.4 -brd bw1098obc` the result is the same.

The calibration process starts and takes a picture of the chess grid. Eventually it segfaults. I'm concerned with all those write failures. Here is the output:

```
$ ./calibrate.py -s 2.4 
No calibration file. Using Calibration Defaults.
Depthai version required:   0.4.0.0
Compiled mode found: compiled for 7 shaves, 7 cmx_slices and 1 NN_engines 
Using Arguments= {'count': 1, 'square_size_cm': 2.4, 'image_op': 'modify', 'mode': ['capture', 'process'], 'config_overwrite': None, 'board': None, 'field_of_view': 71.86, 'baseline': 9.0, 'swap_lr': True, 'dev_debug': None, 'invert_v': False, 'invert_h': False}
Starting image capture. Press the [ESC] key to abort.
Will take 13 total images, 1 per each polygon.
XLink initialized.
Sending internal device firmware
Successfully connected to device.
Usb speed : High/480Mbps
Mx serial id : 14442C10A193F4D000
Loading config file
Attempting to open stream config_d2h
watchdog started 
Successfully opened stream config_d2h with ID #0!
Closing stream config_d2h: ...
Closing stream config_d2h: DONE.
EEPROM data: valid (v5)
  Board name     : BW1098OBC
  Board rev      : R0M0E0
  HFOV L/R       : 71.86 deg
  HFOV RGB       : 68.7938 deg
  L-R   distance : 7.5 cm
  L-RGB distance : 3.75 cm
  L/R swapped    : yes
  L/R crop region: center
  Rectification Rotation R1 (left):
    0.999733,   -0.022838,    0.003520,
    0.022827,    0.999735,    0.003021,
   -0.003588,   -0.002940,    0.999989,
  Rectification Rotation R2 (right):
    0.999603,   -0.028019,   -0.003088,
    0.028010,    0.999603,   -0.003024,
    0.003172,    0.002937,    0.999991,
  Calibration intrinsic matrix M1 (left):
  854.401672,    0.000000,  635.667603,
    0.000000,  854.906799,  407.493988,
    0.000000,    0.000000,    1.000000,
  Calibration intrinsic matrix M2 (right):
[...]

PR #315, opened 2021-01-21
Gen1 'depthai_demo' / 'calibrate' script updates

alex-luxonis:
- Add `-fusb2` option to `calibrate.py` - enabling a workaround for issue https://github.com/luxonis/depthai/issues/305 (cherry-picked the commit mentioned there)
- Remove command-line options that can use `-brd / --board` as a simpler/safer replacement (making sure no option is missed):
    - `-fv  / --field-of-view`
    - `-rfv / --rgb-field-of-view`
    - `-b   / --baseline`
    - `-r   / --rgb-baseline`
    - `-w   / --no-swap-lr`
- Set defaults to BW1098OBC / OAK-D if `-brd` is not specified
- Minor logging cleanup: don't print `Depthai required version` when it matches with `installed`, print just `installed`


szabi-luxonis (approved):
LGTM

PR #316, opened 2021-01-22
Set default mono resolution to 400p; Limit 1st stage cnn choices

szabi-luxonis:
'landmarks-regression-retail-0009', 'facial-landmarks-35-adas-0002', 'emotions-recognition-retail-0003'
are not supposed to be run as first stage networks.
Note: age-gender-recognition can't be run as second stage network due to gen1 limitations. (it has 2 tensor outputs)


alex-luxonis (approved):
LGTM.
Would it make sense to also change the default for `-cnn` to some face detector (instead of the default mobilenet-ssd) if only `-cnn2` is specified? If easy enough?

PR #322, opened 2021-01-25
Allow user to increase window size with mouse drag

matthewaltenburg:


VanDavv (approved):
Tested on my end, works great. Thanks @matthewaltenburg

PR #331, opened 2021-02-03
Update Gen1 depthai: firmware fix for `-fusb2` on Windows

alex-luxonis:
USB descriptor was improperly set up, causing enumeration failure with `-fusb2` / `--force_usb2` on Windows (more strict checks).

It shouldn't affect the behavior on macOS or Linux.

On Linux for example, `dmesg` was just reporting `unable to get BOS descriptor`, but the device worked fine.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,7 +3,7 @@ opencv-python==4.5.1.48
 requests==2.24.0
 argcomplete==1.12.1
 pyyaml==5.3.1
-depthai==0.4.1.1
+# depthai==0.4.1.1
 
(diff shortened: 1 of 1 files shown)


szabi-luxonis (approved):
Please merge into `main` too.

alex-luxonis:
> Please merge into `main` too.

Should we merge as-is to main, or create a release (to have it on pypi)?

szabi-luxonis:
> > Please merge into `main` too.
> 
> Should we merge as-is to main, or create a release (to have it on pypi)?

Yes.

PR #350, opened 2021-04-04
added spatial option (-d)

Erol444:
added spatial option (-d) that displays coordinates of a detected object

### DIFF
--- a/gen2_demo.py
+++ b/gen2_demo.py
@@ -27,6 +27,18 @@ parser.add_argument("-cnn", "--cnn_model", default="mobilenet-ssd", type=str, ch
 parser.add_argument('-sh', '--shaves', default=13, type=int, help="Name of the nn to be run from default depthai repository")
 parser.add_argument('-cnn-size', '--cnn-input-size', default=None, help="Neural network input dimensions, in \"WxH\" format, e.g. \"544x320\"")
 parser.add_argument("-rgbf", "--rgb_fps", default=30.0, type=float, help="RGB cam fps: max 118.0 for H:1080, max 42.0 for H:2160. Default: %(default)s")
+
+def dct_range(arg):
+    try:
+        val = int(arg)
+    except ValueError:
+        raise argparse.ArgumentTypeError("Must be an integer number")
+    if 255 < val or val < 0:
+        raise argparse.ArgumentTypeError("Argument must be between 0 and 255")
+    return val
+parser.add_argument("-dct", "--disparity_confidence_threshold", default=200, type=dct_range, help="Disparity confidence threshold, used for depth measurement. Default: %(default)s")
+parser.add_argument("-med", "--stereo_median_size", default=7, type=int, choices=[0,3,5,7], help="Disparity / depth median filter kernel size (N x N) . 0 = filtering disabled. Default: %(default)s")
+parser.add_argument('-lrc', '--stereo_lr_check', action="store_true", help="Enable stereo 'Left-Right check' feature.")
 args = parser.parse_args()
 
 debug = not args.no_debug
@@ -55,7 +67,10 @@ if args.cnn_input_size is None:
 else:
     in_w, in_h = map(int, args.cnn_input_size.split('x'))
 
-
+if args.stereo_median_size == 0: median = dai.StereoDepthProperties.MedianFilter.MEDIAN_OFF
+elif args.stereo_median_size == 3: median = dai.StereoDepthProperties.MedianFilter.KERNEL_3x3
+elif args.stereo_median_size == 5: median = dai.StereoDepthProperties.MedianFilter.KERNEL_5x5
+elif args.stereo_median_size == 7: median = dai.StereoDepthProperties.MedianFilter.KERNEL_7x7
 class NNetManager:
     source_choices = ("rgb", "left", "right", "host")
     config = None
@@ -212,8 +227,10 @@ def create_pipeline(use_camera, use_hq, use_depth, nn_pipeline=None):
     if use_depth:
         nodes.stereo = p.createStereoDepth()
         nodes.stereo.setOutputDepth(True)
-        nodes.stereo.setConfidenceThreshold(255)
+        nodes.stereo.setConfidenceThreshold(args.disparity_confidence_threshold)
         nodes.stereo.setOutputRectified(True)
+        nodes.stereo.setMedianFilter(median)
+        nodes.stereo.setLeftRightCheck(args.stereo_lr_check)
 
         nodes.mono_left = p.createMonoCamera()
         nodes.mono_left.setResolution(dai.MonoCameraProperties.SensorResolution.THE_400_P)
--- a/gen2_demo.py
+++ b/gen2_demo.py
@@ -216,12 +216,12 @@ def create_pipeline(use_camera, use_hq, use_depth, nn_pipeline=None):
         nodes.stereo.setOutputRectified(True)
 
         nodes.mono_left = p.createMonoCamera()
-        nodes.mono_left.setResolution(dai.MonoCameraProperties.SensorResolution.THE_720_P)
+        nodes.mono_left.setResolution(dai.MonoCameraProperties.SensorResolution.THE_400_P)
         nodes.mono_left.setBoardSocket(dai.CameraBoardSocket.LEFT)
         nodes.mono_left.out.link(nodes.stereo.left)
 
         nodes.mono_right = p.createMonoCamera()
-        nodes.mono_right.setResolution(dai.MonoCameraProperties.SensorResolution.THE_720_P)
+        nodes.mono_right.setResolution(dai.MonoCameraProperties.SensorResolution.THE_400_P)
         nodes.mono_right.setBoardSocket(dai.CameraBoardSocket.RIGHT)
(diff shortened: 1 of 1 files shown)


Erol444:
To note:
- Since rgb-depth alignment is still WIP, I used right mono camera (as suggested by Brandon)
- I wanted to use "high quality" right camera stream but bounding boxes were off (so I commented out that feature) 
- Haven't yet figured out how to add spatial info if NN isn't yolo/mobilenet, thus it has a TODO comment.

Luxonis-Brandon:
Thanks, Erik!  Quite excited for this and for our gen2 demo to become more or less feature parity with gen1 (as this is seeming more and more reasonably tractable).

gen2_demo.py
```diff
@@ -21,16 +21,31 @@
...
                     help="Path to video file to be used for inference (conflicts with -cam)")
 parser.add_argument('-lq', '--lowquality', action="store_true", help="Low quality visualization - uses resized frames")
+parser.add_argument('-d', '--depth', action="store_true", help="Use depth information")
 parser.add_argument('-cnnp', '--cnn_path', type=Path, help="Path to cnn model directory to be run")
 parser.add_argument("-cnn", "--cnn_model", default="mobilenet-ssd", type=str, choices=CNN_choices, help="Cnn model to run on DepthAI")
 parser.add_argument('-sh', '--shaves', default=13, type=int, help="Name of the nn to be run from default depthai repository")
 parser.add_argument('-cnn-size', '--cnn-input-size', default=None, help="Neural network input dimensions, in \"WxH\" format, e.g. \"544x320\"")
 parser.add_argument("-rgbf", "--rgb_fps", default=30.0, type=float, help="RGB cam fps: max 118.0 for H:1080, max 42.0 for H:2160. Default: %(default)s")
+
+def dct_range(arg):
```
VanDavv, line 31: can you change it to the implementation like here - https://github.com/luxonis/depthai-experiments/blob/master/gen2-collecting-training-data/main.py#L11 ? 

This way, we will make a generic range checker for other uses too (and it's already implemented so it's copy-paste'able

gen2_demo.py
```diff
@@ -53,7 +68,10 @@
 else:
     in_w, in_h = map(int, args.cnn_input_size.split('x'))
 
-
+if args.stereo_median_size == 0: median = dai.StereoDepthProperties.MedianFilter.MEDIAN_OFF
```
VanDavv, line 71: I would make this an `else` condition, as default in case `args.stereo_median_size` would get any value other than predicted (e.g. at some place between argparse check and this check the `args.stereo_median_size` would get changed to invalid value, this item is mutable). 
Other approach would be to add an `else` condition with `raise ValueError`. 
Right now, if we would mutate this value to not predicted one, the `median` value would not be defined at all

```suggestion
if args.stereo_median_size == 3: median = dai.StereoDepthProperties.MedianFilter.KERNEL_3x3
elif args.stereo_median_size == 5: median = dai.StereoDepthProperties.MedianFilter.KERNEL_5x5
elif args.stereo_median_size == 7: median = dai.StereoDepthProperties.MedianFilter.KERNEL_7x7
else: median = dai.StereoDepthProperties.MedianFilter.MEDIAN_OFF
```

gen2_demo.py
```diff
@@ -123,6 +151,27 @@ def create_nn_pipeline(self, p, nodes):
             xin.setStreamName(self.input_name)
             xin.out.link(nn.input)
             setattr(nodes, self.input_name, xout)
+        elif self.source == "right": # Use spatial information
+            nodes.xout_right = p.createXLinkOut()
+            nodes.xout_right.setStreamName("right")
```
VanDavv, line 156: I would separate this change from this function. This one is meant to serve only as a nn creator, so it should only have calls related to creating NN. 
Right here, both right xout and image manip are created, which should be moved to a different function.
The flow should look more like

- user provides params
- we resolve these params to what we need (so if he wants depth, we need to enable left and right camera flags, also the depth flag. In case we would like also nn, we also enable right_nn_camera flag and create nn with source=right.
- then, we do a pipeline functions based on a calls, so it would look like
  ```
  if enable_left_cam:
      pipelineManager.createLeftCam()
  if enable_right_cam:
      pipelineManager.createRightCam()
  ```
My fear is that we will get quickly tighten up by putting up too much things in a single place

gen2_demo.py
```diff
@@ -171,12 +220,29 @@ def fps(self):
         return self.frame_cnt / (self.timestamp - self.start)
 
 
-def create_pipeline(use_camera, use_hq, nn_pipeline=None):
+def create_pipeline(use_camera, use_hq, use_depth, nn_pipeline=None):
```
VanDavv, line 223: Seeing how much this function has grown, can we make it like a `PipelineManager`?
that would be a class with functions like `createDepth` or `createLeftCam` so that we can make it more flexible?
This relates also to the comment above about separating the logic in this script

VanDavv (commented):
I think that my comments are not so much related to changes made here, but to overall approach that is taken in this script, that as it grows, I see that some parts of the code might have been written in a more flexible way.
So @Erol444 feel free to ignore some of them, as resolving these might be a subject of another PR. 
I gave it a try and it works great, with one minor change to rename the window as we display `right` camera image, not `rgb`
image

gen2_demo.py
```diff
@@ -123,6 +156,20 @@ def create_nn_pipeline(self, p, nodes):
             xin.setStreamName(self.input_name)
             xin.out.link(nn.input)
             setattr(nodes, self.input_name, xout)
+        elif self.source == "right": # Use spatial information
+            # Set XLinkOut sources
+            nn.passthrough.link(nodes.xout_right.input)
```
VanDavv, line 161:
```suggestion
            nn.passthrough.link(nodes.xout_right.input)  # TODO add --sync flag support to determine if we want a passthrough of just right in xout_right
```

gen2_demo.py
```diff
@@ -171,49 +218,102 @@ def fps(self):
...
+        if 'mono_left' not in vars(self.nodes): self.create_left_cam()
+        if 'mono_right' not in vars(self.nodes): self.create_right_cam()
+
+        self.nodes.mono_left.out.link(self.nodes.stereo.left)
+        self.nodes.mono_right.out.link(self.nodes.stereo.right)
+
+    def create_left_cam(self):
+        self.nodes.mono_left = self.p.createMonoCamera()
+        self.nodes.mono_left.setResolution(dai.MonoCameraProperties.SensorResolution.THE_400_P)
+        self.nodes.mono_left.setBoardSocket(dai.CameraBoardSocket.LEFT)
```
VanDavv, line 257: maybe we could also add xouts here?

gen2_demo.py
```diff
@@ -171,49 +218,102 @@ def fps(self):
...
+    def create_depth(self, dct, median, lr):
+        self.nodes.stereo = self.p.createStereoDepth()
+        self.nodes.stereo.setOutputDepth(True)
+        self.nodes.stereo.setOutputRectified(True)
+        self.nodes.stereo.setConfidenceThreshold(dct)
+        self.nodes.stereo.setMedianFilter(median)
+        self.nodes.stereo.setLeftRightCheck(lr)
+
+        # Create mono left/right cameras if we haven't already
+        if 'mono_left' not in vars(self.nodes): self.create_left_cam()
```
VanDavv, line 248:
```suggestion
        if not hasattr(self.nodes, 'mono_left'): self.create_left_cam()
```

gen2_demo.py
```diff
@@ -171,49 +218,102 @@ def fps(self):
...
+            self.nodes.manip = self.p.createImageManip()
+            self.nodes.manip.initialConfig.setResize(in_w, in_h)
+            # The NN model expects BGR input. By default ImageManip output type would be same as input (gray in this case)
+            self.nodes.manip.initialConfig.setFrameType(dai.RawImgFrame.Type.BGR888p)
+            # Set the stereo node's rectifiedRight stream as the ImageManip's source
+            self.nodes.stereo.rectifiedRight.link(self.nodes.manip.inputImage)
+
+            # Create an XLinkOut node through which we will send right_camera stream
+            self.nodes.xout_right = self.p.createXLinkOut()
+            self.nodes.xout_right.setStreamName("right")
```
VanDavv, line 276: I think this should be moved to createRightCam

gen2_demo.py
```diff
@@ -171,49 +218,102 @@ def fps(self):
...
+
+            if nn_manager.sbb:
+                # If we want to display spatial bounding boxes, create XLinkOut node for:
+                # a) Spatial bounding boxes
+                self.nodes.xout_sbb = self.p.createXLinkOut()
+                self.nodes.xout_sbb.setStreamName("sbb")
```
VanDavv, line 282: IMO this should go to NN manager function

gen2_demo.py
```diff
@@ -171,49 +218,102 @@ def fps(self):
...
+                # b) depth map
+                self.nodes.xout_depth = self.p.createXLinkOut()
+                self.nodes.xout_depth.setStreamName("depth")
```
VanDavv, line 285: And this to depth creation function

VanDavv (changes requested):

VanDavv (approved):
LGTM, thanks @Erol444!

PR #359, opened 2021-04-25
Gen2 code refactor + improvements

VanDavv:
This PR improves the code structure making it more flexible, also reorganizes where certain logic is kept in the code.

Contains also various runtime fixes to make the script more usable and stable

### DIFF
--- a/depthai_demo.py
+++ b/depthai_demo.py
@@ -345,6 +345,7 @@ with dai.Device(pm.p) as device:
     fps = FPSHandler() if conf.useCamera else FPSHandler(cap)
     seq_num = 0
     frame = None
+    depth_frame = None
     detections = []
     # Spatial bounding box ROIs (region of interests)
     color = (255, 255, 255)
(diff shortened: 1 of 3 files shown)


Luxonis-Brandon:
Just one note: We could likely automatically detect the number of cameras connected.  

image

And is there a reason that the NN FPS is only 27FPS?  How many SHAVES/NCE are being used here?  This seems a bit low.

image

Also we should add the disparity confidence slider that was in Gen1.  That is quite nice to have.

Luxonis-Brandon:
Oh but yes, the comments above definitely can go into a future PR/etc. @VanDavv .  I think it was the right call to have merged this.

VanDavv:
@Luxonis-Brandon I merged this because it runs more stable and flexible than the version on the gen2 branch (to which this branch was merged).
@themarpe added the capability to detect connected cameras here - https://github.com/luxonis/depthai-python/tree/device_improvements - so will give it a try now
(posting here for clarity)

Luxonis-Brandon:
Sounds great thanks!  And agreed.  :-)

PR #361, opened 2021-04-29
Gen2 timestamp fixes

szabi-luxonis:
Fixed [division by zero ](https://discord.com/channels/790680891252932659/799407361986658354/837100723690536983) error, switched to monotonic timer.


VanDavv (commented):
Thanks for these fixes!

PR #362, opened 2021-04-30
Gen2 device detection

VanDavv:


depthai_demo.py
```diff
@@ -308,33 +307,34 @@ def create_nn(self):
...
 
-nn_manager = NNetManager(
-    model_name=conf.getModelName(),
-    model_dir=conf.getModelDir(),
-    source=conf.getModelSource(),
-    use_depth=conf.useDepth,
-    use_hq=conf.useHQ
-)
+# Pipeline is defined, now we can connect to the device
+with dai.Device() as device:
```
themarpe, line 311: If you know the intended OpenVINO version rather use that explicitly - eg if you use 2021.3 for networks, specify this here. (Otherwise latest is used by default, which could throw of things if new version is added)

PR #377, opened 2021-05-19
Add support for custom NN models

VanDavv:
With this PR, users will be able to add custom neural network support without digging into the main depthai_demo.py source code.

A deeplabv3p_person model is used to show an integration example

image

Changes added:
- NN input size is fetched from model json config file (under `nn_config.input_size`)
- Added a new json config option - `handler` - that specified a path to file that should handle both decoding and display of the NN results. Each handler file needs to contain both `decode` and `draw` methods, they can also import the necessary modules.


themarpe (changes requested):
Would the capability to query blob details be more appropriate here? 
I can try spinning it up if no time pressure on the PR?

VanDavv:
Yes, it would be very helpful @themarpe. Maybe we can merge it now and then improve when the API is available?

themarpe (approved):
Lets go for it - I'm not sure on my timing regarding the feature

Erol444 (approved):

PR #378, opened 2021-05-20
Add custom callbacks support, allowing to customize demo script behaviour

VanDavv:
This PR implements support for custom callbacks that users can add in order to customize the behavior of the demo script.

For example, if a user would like to add a label to each frame, that would indicate its origin, he can do it now without digging into a code, but modifying the `on_show_frame` method inside `callbacks.py` into the following

```python
from cv2 import cv2

# ...

def on_show_frame(frame, source):
    cv2.rectangle(frame, (frame.shape[1] - 100, frame.shape[0] - 30), (frame.shape[1], frame.shape[0]), (255, 255, 255), cv2.FILLED)
    cv2.putText(frame, source, (frame.shape[1] - 80, frame.shape[0] - 10), cv2.FONT_HERSHEY_TRIPLEX, 0.5, (0, 0, 0))

```

Which will result in the following output

image

Supported callbacks:

- `on_new_frame(frame, source)` - called when a new frame arrives from the device
- `on_show_frame(frame, source)` - called when a frame is about to be displayed
- `on_nn(nn_packet)` - called when a new nn packet is received
- `on_report(report)` - called when a new usage report is generated
- `on_setup(**kwargs)` - called when the demo script finished it's setup, with`kwargs` dictionary containing all variables inside the script. So, if someone would like e.g. to modify some attributes in PreviewManager, he can access it either using `kwargs["pv"]` (as `pv` is the variable name assigned to PreviewManager) or by changing the function definition to `on_setup(pv, **kwargs)` and use the `pv` parameter
- `on_teardown(**kwargs)` - called when the demo script is about to terminate
- `on_iter(**kwargs)` - called each time when a new iteration of the script's internal loop is started

Usage of `**kwargs` and `locals()` allows to pick just about anything from the script, so there is a very high degree of freedom in potential modifications

### DIFF
--- a/callbacks.py
+++ b/callbacks.py
@@ -1,0 +1,26 @@
+def on_new_frame(frame, source):
+    pass
+
+
+def on_show_frame(frame, source):
+    pass
+
+
+def on_nn(nn_packet):
+    pass
+
+
+def on_report(report):
+    pass
+
+
+def on_setup(**kwargs):
+    pass
+
+
+def on_teardown(**kwargs):
+    pass
+
+
+def on_iter(**kwargs):
+    pass
--- a/depthai_demo.py
+++ b/depthai_demo.py
@@ -10,13 +10,12 @@ import importlib.util
 import cv2
 import depthai as dai
 import numpy as np
-import callbacks
 from depthai_helpers.version_check import check_depthai_version
 import platform
 
 from depthai_helpers.arg_manager import parse_args
 from depthai_helpers.config_manager import BlobManager, ConfigManager
-from depthai_helpers.utils import frame_norm, to_planar, to_tensor_result
+from depthai_helpers.utils import frame_norm, to_planar, to_tensor_result, load_module
 
 print('Using depthai module from: ', dai.__file__)
 print('Depthai version installed: ', dai.__version__)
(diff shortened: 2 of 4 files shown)


themarpe (changes requested):
With `depthai` taking a part of middleware for our library I think we must stride towards "depthai_demo" becoming an program capable of executing "plugins/apps". These apps would be very simple to write (if little modifications are needed) and wouldn't require the verbosity of depthai library.

Outlook:
 - depthai (repository)
   - app1 (example/...)
   - app2 (example/...)
 
And users could then do: 
- myApp (Users custom application)
- depthai (repository)

depthai --app myApp

Thats my vision for further improvements.

Short term however, I see similar with the callbacks. Instead of file being a part of source tree, lets make this only a template which users can copy and then pass as a parameter:
```
depthai_demo.py --callback my_callback.py
```

This allows for following:
 - my_callback
 - my_callback2
 - depthai

While not needing to have 2 depthai repositories cloned.

TLDR (IMO): `callbacks.py` -> `callback_template.py` & argument for specifying the actual callback file.

What do you think?

VanDavv:
@themarpe so let's maybe split this into stages:

- I would add support in this PR for the custom `--callback` file, so that users will be able to have multiple of these. However I would leave the default `callbacks.py` file as then we don't need to explain anything more than `just edit this file and you're ready to go` if someone wants to give it a try. Also, users can use this file as a template for their callback files.

- Add a ticket and discuss the `--app` param further, that would be the next level of customization (but since it will enable us to migrate some of the existing examples here as apps, l would add it rather sooner than later)

themarpe:
Sounds good - one thing that I'd like to do is to remove the need for users to actually modify the "source" of the `depthai`
I guess if both options are available then thats okay, although it might be better to encourage having it separate for customizations. (allows for users to only receive the `depthai_demo.py` as a program).

Lets go with 1. for now as you mentioned and discuss more for 2. :+1:

themarpe (approved):

PR #384, opened 2021-05-24
Throw error when accessing Gen1-based scripts (will be migrated later)

VanDavv:
Some of the scripts we still had in our repository were not working after migration to Gen2. This PR removes them, not to cause confusion for new users, but they will be added back once we make them work with the new script


depthai_helpers/calibration_utils.py
```diff
@@ -1,868 +0,0 @@
-import cv2
```
alex-luxonis, line 1: We shouldn't remove this file, as it doesn't contain DepthAI specific stuff. It is imported by `calibrate.py` - and I think removing that isn't good either, as users would wonder where it disappeared. Instead I think we should print a warning message (and then exit) inside `calibrate.py` that is wasn't migrated to Gen2, and if users want to run the calibration, they should switch to the `gen1_main` branch.

VanDavv: @alex-luxonis it contained 
```
import consts.resource_paths
```
which we don't have anymore. But I agree, printing a message or throwing an error instead of removing the file will be better for UX

VanDavv:
@alex-luxonis per your suggestion, instead of deleting the calibration files, the error is thrown with the following content

```
  File "/home/vandavv/dev/luxonis/depthai/calibrate.py", line 3, in <module>
    raise NotImplementedError("Calibration is not yet available in Gen2 demo. To calibrate your device, switch to \"gen1_main\" branch and try again")
NotImplementedError: Calibration is not yet available in Gen2 demo. To calibrate your device, switch to "gen1_main" branch and try again
```

szabi-luxonis:
> @alex-luxonis per your suggestion, instead of deleting the calibration files, the error is thrown with the following content
> 
> ```
> Traceback (most recent call last):
>   File "/home/vandavv/dev/luxonis/depthai/calibrate.py", line 3, in <module>
>     raise NotImplementedError("Calibration is not yet available in Gen2 demo. To calibrate your device, switch to \"gen1_main\" branch and try again")
> NotImplementedError: Calibration is not yet available in Gen2 demo. To calibrate your device, switch to "gen1_main" branch and try again
> ```

When merging select the option "Squash and merge" otherwise history will be messed up due to removing/readding of files.
Or better: soft reset and force push again on this branch, it's cleaner:
```
git soft --reset HEAD~2
git commit -m "..."
git push --force origin HEAD:remove_gen1_files
```

VanDavv:
@szabi-luxonis done

szabi-luxonis (approved):

PR #385, opened 2021-05-26
Add support for precompiled blob file and custom OpenVINO version

VanDavv:
With this PR, users will be able to use `--openvino_version` flag to specify OpenVINO version used in pipeline.
Additionally, users are now able to run depthai_demo with custom blob, without blobconverter


Erol444 (approved):
LGTM, thanks:)

PR #388, opened 2021-05-27
Gen2 calibration

saching13:
- Enables calibration of Stereo setup along with or without RGB camera
- Option to swap Left and Right camera
- choose Lens Position of RGB camera during calibration
- Create Mesh files
- Write to EEPROM

Added the Default Charuco board needed.


Luxonis-Brandon:
Exciting!

Luxonis-Brandon (approved):
My opinion doesn't really count on this review.  But just glancing through at the highest-possible level this LGTM.

calibrate.py
```diff
@@ -1,39 +1,55 @@
...
 import shutil
-import json
-from depthai_helpers.version_check import check_depthai_version
-check_depthai_version()
+from datetime import datetime
+from argparse import ArgumentParser
+import argparse
+
+import depthai as dai
+from depthai_helpers.calibration_utils import *
```
VanDavv, line 22:
```suggestion
import depthai_helpers.calibration_utils as utils
```

importing using `*` can shadow names from this file. Using named import will prevent it and IMO it would be easier to understand where a function is defined

calibrate.py
```diff
@@ -135,103 +128,128 @@ class Main:
 
     def __init__(self):
         self.args = vars(parse_args())
```
VanDavv, line 130:
```suggestion
        self.args = parse_args()
```
Removing `vars` will allow to access parameters like `self.focus_value = self.args.rgbLensPosition` instead of `self.focus_value = self.args['rgbLensPosition']`

calibrate.py
```diff
@@ -135,103 +128,128 @@ class Main:
...
-                },
-        }
+        self.aruco_dictionary = cv2.aruco.Dictionary_get(
+            cv2.aruco.DICT_4X4_1000)
+        self.focus_value = self.args['rgbLensPosition']
         if self.args['board']:
             board_path = Path(self.args['board'])
             if not board_path.exists():
-                board_path = Path(consts.resource_paths.boards_dir_path) / Path(self.args['board'].upper()).with_suffix('.json')
+                board_dir = str(
```
VanDavv, line 137: can be simplified

```
board_path = (Path(__file__).parent / 'resources/boards' / self.args['board'].upper()).with_suffix('.json').resolve()
```

depthai_helpers/calibration_utils.py
```diff
@@ -144,615 +176,749 @@ def calibrate(self, filepath, square_size, out_filepath, flags):
...
-        # Arrays to store object points and image points from all the images.
+
+    def parse_frame(self, frame, stream_name, file_name):
+        file_name += '.png'
+        # filename = image_filename(stream_name, self.current_polygon, self.images_captured)
+        # print(self.data_path + "/{}/{}".format(stream_name, file_name))
+        ds_path = self.data_path + "/{}".format(stream_name)
+        print(ds_path)
+        if not os.path.exists(ds_path):
+            os.makedirs(ds_path)
```
VanDavv, line 197: can be simplified

```
ds_path = (Path(self.data_path) / stream_name).with_suffix('.png')
ds_path.mkdir(parents=True, exists_ok=True)
```

VanDavv (commented):
I found just a couple of minor changes I could catch, but other than that LGTM - I'm very impressed by the size and complexity of this PR, kudos @saching13!

calibrate.py
saching13: Oohh. That's why I was not able to do that one. Thanks

depthai_helpers/calibration_utils.py
saching13: I removed this. There is no use for this.

saching13:
This is ready for merge when the current Develop goes to release.

VanDavv (approved):

PR #389, opened 2021-05-31
Add custom_model example

VanDavv:
This PR is related to https://github.com/luxonis/depthai-docs-website/pull/198

Adds example files that allow to reproduce steps in the docs


Erol444 (approved):
LGTM, thanks:)

alex-luxonis:
Should we store this on Artifactory instead?
image

Would need to have a way to download it, so not as clean, but just to not bloat the repo again.

VanDavv:
@alex-luxonis I tried to pick a small blob but wanted to have it available right away and working without any additional steps. My fear was that adding custom blobs could be complicated if doing for the first time, so wanted to make it easier by showing our users a working example

PR #396, opened 2021-06-10
Add tweaks and error fixes in calibrate.py

VanDavv:
- Fixes error when calling calibration script without `dataset` directory
- Hides some debug prints behind `debug` flag
- Removes old pattern files and adds charuco in PNG


Erol444 (approved):

calibrate.py
```diff
@@ -582,13 +586,14 @@ def calibrate(self):
     def run(self):
         if 'capture' in self.args.mode:
             try:
-                # if self.args.imageOp == 'delete':
-                shutil.rmtree('dataset/')
+                if self.args.imageOp == 'delete' and Path('dataset').exists():
```
saching13, line 589: I was thinking of deleting the dataset the old by default when they start a new calibration procedure by default.

saching13: This would work better in case if the users had messed up with files inside the folder. like adding additional images inside which will get used in calibration in the new execution unless deleted.

VanDavv: Sure, should we remove the `imageOp` then?

saching13: Yup.

saching13:
Can you change that `-ms` and `-s` required = True and remove the default and it should be good to merge.

And Thanks for making these changes.

saching13 (approved):

VanDavv:
Ok, I removed the defaults and `imageOp` param. Thanks all!

Issue #397, opened 2021-06-10
openpose2 installing


R-abodyak:
hi ,
when I write $ python depthai_demo.py -cnn openpose2
in the command

I get this problem
"exit_code": 1,
"message": "Command failed with exit code 1, command: /opt/intel/openvino/deployment_tools/inference_engine/lib/intel64/myriad_compile -m /tmp/blobconverter/feea082f408c45eeae0d62e9c8360f42/openpose2/FP16/openpose2.xml -o /tmp/blobconverter/feea082f408c45eeae0d62e9c8360f42/openpose2/FP16/openpose2.blob -c /tmp/blobconverter/feea082f408c45eeae0d62e9c8360f42/myriad_compile_config.txt -ip U8",
"stderr": "The support of IR v5 has been removed from the product. Please, convert the original model using the Model Optimizer which comes with this version of the OpenVINO to generate supported IR version.\n",
"stdout": "Inference Engine: \n\tAPI version ............ 2.1\n\tBuild .................. 2021.3.0-2787-60059f2c755-releases/2021/3\n\tDescription ....... API\n"
}
Traceback (most recent call last):
File "C:\Users\HI TECH\Desktop\depthai\depthai_demo.py", line 574, in
nn_manager = NNetManager(
File "C:\Users\HI TECH\Desktop\depthai\depthai_demo.py", line 198, in init
self.blob_path = BlobManager(model_dir=self.model_dir, model_name=self.model_name).compile(conf.args.shaves)
File "C:\Users\HI TECH\Desktop\depthai\depthai_helpers\config_manager.py", line 313, in compile
self.blob_path = blobconverter.compile_blob(
File "C:\Users\HI TECH\AppData\Roaming\Python\Python39\site-packages\blobconverter_init_.py", line 252, in compile_blob
response.raise_for_status()
File "C:\Users\HI TECH\AppData\Roaming\Python\Python39\site-packages\requests\models.py", line 941, in raise_for_status
raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 400 Client Error: BAD REQUEST for url: http://luxonis.com:8080/compile?version=2021.3

what should I do ?

Luxonis-Brandon:
I think I saw this issue raised elsewhere and then closed.  Are you still experiencing the problem?

R-abodyak:
yes . I still stuck

Luxonis-Brandon:
[...]

PR #398, opened 2021-06-14
Add openvino_version to model config & add openpose2 handler

VanDavv:
With this PR, the download issue from #397 will be resolved

The download failed on `main` due to the following error:

```
The support of IR v5 has been removed from the product. Please, convert the original model using the Model Optimizer which comes with this version of the OpenVINO to generate supported IR version
```

We should update the openpose IR files, but I decided to add an ability to specify a custom version in the model config file and utilize our blobconverter's mechanism to download a blob in a specific OpenVINO version.
This way, this model is now downloaded by default using OpenVINO 2020.4.

### DIFF
--- a/depthai_helpers/utils.py
+++ b/depthai_helpers/utils.py
@@ -26,11 +26,11 @@ def to_tensor_result(packet):
     data = {}
     for tensor in packet.getRaw().tensors:
         if tensor.dataType == dai.TensorInfo.DataType.INT:
-            data[tensor.name] = np.array(packet.getLayerInt32(tensor.name)).reshape(tensor.dims)
+            data[tensor.name] = np.array(packet.getLayerInt32(tensor.name)).reshape(tensor.dims[::-1])
         elif tensor.dataType == dai.TensorInfo.DataType.FP16:
-            data[tensor.name] = np.array(packet.getLayerFp16(tensor.name)).reshape(tensor.dims)
+            data[tensor.name] = np.array(packet.getLayerFp16(tensor.name)).reshape(tensor.dims[::-1])
         elif tensor.dataType == dai.TensorInfo.DataType.I8:
-            data[tensor.name] = np.array(packet.getLayerUInt8(tensor.name)).reshape(tensor.dims)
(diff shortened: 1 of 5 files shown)


depthai_helpers/utils.py
```diff
@@ -26,11 +26,11 @@ def to_tensor_result(packet):
...
     for tensor in packet.getRaw().tensors:
         if tensor.dataType == dai.TensorInfo.DataType.INT:
-            data[tensor.name] = np.array(packet.getLayerInt32(tensor.name)).reshape(tensor.dims)
+            data[tensor.name] = np.array(packet.getLayerInt32(tensor.name)).reshape(tensor.dims[::-1])
         elif tensor.dataType == dai.TensorInfo.DataType.FP16:
-            data[tensor.name] = np.array(packet.getLayerFp16(tensor.name)).reshape(tensor.dims)
+            data[tensor.name] = np.array(packet.getLayerFp16(tensor.name)).reshape(tensor.dims[::-1])
         elif tensor.dataType == dai.TensorInfo.DataType.I8:
-            data[tensor.name] = np.array(packet.getLayerUInt8(tensor.name)).reshape(tensor.dims)
+            data[tensor.name] = np.array(packet.getLayerUInt8(tensor.name)).reshape(tensor.dims[::-1])
```
VanDavv, line 33: CC @themarpe on this, `tensor.dims` is returning tensor dimensions but in reverse order. For example, for `openpose2`, the output dimensions are
```
[1, 57, 46, 54]
```
whereas the `tensor.dims` returns
```
[54, 46, 57, 1]
```

It had no impact in deeplabv3p_person since it's output looks like
```
[256, 256, 1, 1]  # incorrect, current
[1, 1, 256, 256]  # expected
```

so after squashing it still ended up being `[256, 256]`

Added reverse here (`[::-1]`) but I think we should fix that in the API too

themarpe: Resolved in https://github.com/luxonis/depthai-core/pull/160
Either do a version check and use reverse if < 2.6.0, or we can update later on?

VanDavv: I'd say to update it later on when we'll be bumping 2.6.0

Erol444 (approved):
LGTM, thanks:)

PR #399, opened 2021-06-16
Show different colors per detected label

VanDavv:
With this PR, each label will have assigned a different color with which it will be drawn on the preview windows

image


Erol444 (approved):
Thanks:)

Luxonis-Brandon:
Thanks!

PR #400, opened 2021-06-17
Added object counting feature

Erol444:
`python3.6 depthai_demo.py --count 5`
image

`python3.6 depthai_demo.py --count tvmonitor`
image

I'm using bottles as an example because the model detects them accurately. 
Any ideas on how to get multiples of the object? So it would display eg. `2 people` instead of `2 person`

### DIFF
--- a/depthai_demo.py
+++ b/depthai_demo.py
@@ -226,6 +226,9 @@ class NNetManager:
         if conf.args.cnn_input_size:
             self.input_size = tuple(map(int, conf.args.cnn_input_size.split('x')))
 
+        # Count objects detected on the frame
+        self.count = conf.getCountLabel(self)
+
     @property
     def should_flip_detection(self):
         return self.source in ("rectified_left", "rectified_right") and not conf.args.stereo_lr_check
@@ -261,7 +264,7 @@ class NNetManager:
         else:
             # TODO use createSpatialLocationCalculator
             nn = p.createNeuralNetwork()
-        
+
         self.blob_path = self.blob_manager.compile(conf.args.shaves, self.openvino_version)
         nn.setBlobPath(str(self.blob_path))
         nn.setNumInferenceThreads(2)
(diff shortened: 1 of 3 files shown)


Luxonis-Brandon:
Looks great!

depthai_demo.py
```diff
@@ -226,6 +226,9 @@ def __init__(self, source, model_dir=None, model_name=None):
         if conf.args.cnn_input_size:
             self.input_size = tuple(map(int, conf.args.cnn_input_size.split('x')))
 
+        # Count objects detected on the frame
+        self.count = conf.getCountLabel(self)
```
VanDavv, line 230: Can we rename the self.count and conf.args.count to `count_label`? 

The cli param can stay the same, adding `dest` flag to argparse should make it work, it's only a matter of knowing that does a count variable mean

depthai_demo.py
```diff
@@ -379,6 +382,21 @@ def draw_detection(frame, detection):
...
                 else:
                     draw_detection(source, detection)
+
+            if self.count is not None:
+                def draw_cnt(frame, cnt):
+                    cv2.rectangle(frame, (0, 35), (120, 50), (255, 255, 255), cv2.FILLED)
+                    cv2.putText(frame, f"{self.count}: {cnt}", (5, 46), fps_type, 0.5, fps_color)
+
+                # Count the number of detected objects
+                cnt_iter = filter(lambda x: self.get_label_text(x.label) == self.count, decoded_data)
```
VanDavv, line 392:
```suggestion
                cnt_iter = list(filter(lambda x: self.get_label_text(x.label) == self.count, decoded_data))
```

depthai_demo.py
```diff
@@ -379,6 +382,21 @@ def draw_detection(frame, detection):
...
+                cnt = sum(1 for _ in cnt_iter)
```
VanDavv, line 393:
```suggestion
                cnt = len(cnt_iter)
```

depthai_demo.py
VanDavv: And it would also change it's name from `cnt_iter` to `cnt_list` or other

depthai_helpers/arg_manager.py
```diff
@@ -84,4 +84,5 @@ def parse_args():
                         help="Mono cam fps: max 60.0 for H:720 or H:800, max 120.0 for H:400. Default: %(default)s")
     parser.add_argument('-cb', '--callback', type=Path, default=project_root / 'callbacks.py', help="Path to callbacks file to be used. Default: %(default)s")
     parser.add_argument("--openvino_version", type=str, choices=openvino_versions, help="Specify which OpenVINO version to use in the pipeline")
+    parser.add_argument("--count", type=str, help="Count and display the number of specified objects on the frame. You have to specify the object (label) you want to count.")
```
VanDavv, line 87: I would add that the label can be either a label id or label name (if defined labels in nn config file)

VanDavv (changes requested):

VanDavv (approved):

PR #401, opened 2021-06-18
Add ability to force USB2 mode

VanDavv:


themarpe (commented):
Soon there is going to be any USB speed API capability.
Just something to be aware of, we might rather redo this argument to something more "generic" which can cover all speeds (usb2, gen1, gen2).

And we only use it now for setting usb2, but later just extend for gen1 and gen2

VanDavv:
@themarpe the param is now called `-usbs / --usb_speed` and accepts either `usb2` or `usb3` (default)

themarpe (approved):

VanDavv:
Related docs update - https://github.com/luxonis/depthai-docs-website/pull/217

PR #402, opened 2021-06-18
Enable device flag allowing to list all devices and select a specific one

VanDavv:
allows to call

```
$ depthai_demo.py -dev list
Using depthai module from:  /home/vandavv/dev/luxonis/venvs/demo/lib/python3.8/site-packages/depthai.cpython-38-x86_64-linux-gnu.so
Depthai version installed:  2.5.0.0
Available devices:
[0] 14442C10A182BFD200 [X_LINK_UNBOOTED]
[1] 14442C1061EBAACE00 [X_LINK_UNBOOTED]
```
to list all devices or run the demo script on a specific device

```
$ depthai_demo.py -dev 14442C10A182BFD200
```


Luxonis-Brandon:
This will be nice to have - thanks.

themarpe:
@VanDavv only a nitpick but what if we rather do: `--list / -l` for this operation as its very common across all CLI programs.
And also then no need to special parse the `-dev` argument for this particular string.

VanDavv:
@themarpe it makes no difference to me, but I would name it then `--dev-list` to be more specific about what does the param do. Although in this form it's pretty much the same as `--dev list` so basically a matter if we want to have a space or a dash there :slightly_smiling_face:

themarpe:
@VanDavv the param name being more descriptive here IMO doesn't help as once you have to checkup the actual param which to use for certain action, it doesn't matter that much anymore (from terms of speed / ease of use). 
On the other hand, just trying out something which works almost everywhere else, is much faster.

But it might not be the best action for this parameter, maybe a list of examples/functionality/... would be better?

Either way, using `-l` I'd be nice in an appropriate case. Not sure which one would be more suitable.

I guess we can leave it as is for now

themarpe (approved):

PR #406, opened 2021-06-21
Add the ability to encode (record) camera output

VanDavv:
This PR adds two new flags:

- `-enc / --encode` flag that accepts the camera names to be recorded and also can contain their FPS (defaults to 30). You can pass either just `-enc left` (encode left camera with 30 FPS) or `-enc left right` (encode left and right camera with 30 FPS) or `-enc color left,10 right,10` (encode color camera with 30 FPS and left / right cameras with 10 FPS)
- `-encout / --encode_output` specifies path to the output directory where the encoded files will be stored (in a form of `<camera_name>.h264` )
```
$ ./depthai_demo.py --help
[...]
  -enc ENCODE [ENCODE ...], --encode ENCODE [ENCODE ...]
                        Define which cameras to encode (record) 
                        Format: camera_name or camera_name,enc_fps 
                        Example: -enc left color 
                        Example: -enc color right,10 left,10
  -encout ENCODE_OUTPUT, --encode_output ENCODE_OUTPUT
                        Path to directory where to store encoded files. Default: /home/vandavv/dev/luxonis/depthai
```

### DIFF
--- a/depthai_demo.py
+++ b/depthai_demo.py
@@ -552,6 +552,44 @@ class PipelineManager:
             self.nodes.xout_right.setStreamName(Previews.right.name)
             self.nodes.mono_right.out.link(self.nodes.xout_right.input)
 
+    def create_encoder(self, camera_name, enc_fps):
+        allowed_sources = [Previews.left.name, Previews.right.name, Previews.color.name]
+        if camera_name not in allowed_sources:
+            raise ValueError("Camera param invalid, received {}, available choices: {}".format(camera_name, allowed_sources))
+        node_name = camera_name.lower() + '_enc'
+        xout_name = node_name + "_xout"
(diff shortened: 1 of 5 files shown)


themarpe (changes requested):
Preferably we could save directly to mp4/etc.. in depthai with a simpler library that just writes the needed container before the h26x stream - but the following might also be a good option to run at the end https://ffmpy3.readthedocs.io/en/latest/examples.html

Not sure about crossplatform support though.

So mostly just getting even more UX here as it completely appropriate place to not require any additional users work

VanDavv:
@themarpe I've added it in the recent commit.
This library is overall nice (and seems to be supported also on Windows and OSX), I had to patch it thought for it to work - https://github.com/wchill/ffmpy3/pull/5

Erol444 (approved):
LGTM

PR #408, opened 2021-06-22
Refactor demo code to extract managers to a separate file

VanDavv:
This PR splits the managers from the demo code itself. Some minor fixes included, like constant detection flip when using `rectified_left` or `rectified_right` streams


Erol444 (approved):
Thanks! This is much better. And if the manager file gets too big (it's already 500 lines) we can later just move these managers each in their own file.

PR #409, opened 2021-06-22
Add default board parameter to calibration script

VanDavv:
Since we can calculate the `-ms` param based on `-s`, but only when we're using charuco board from our repository / calib.io, this PR introcudes the `-db` / `--defaultBoard` parameter that enables automatic `-ms` calculation


Erol444 (approved):
LGTM

PR #412, opened 2021-06-24
increase disparity confidence threshold

VanDavv:


Erol444 (approved):
This will be much better for the users, thanks:)

PR #413, opened 2021-06-27
Improve depth quality and allow real-time configuration

VanDavv:
With this PR:
- new sliders were added below depth, disparity and disparity_color previews. They can adjust disparity confidence threshold, bilateral sigma and left right check threshold. Trackbars are in sync, so if we update e.g. the confidence on one slider, other will get updated too
- on the same previews, a new box was added specifying the current median filter used. Users can use a different filter with`m` key. Future dynamic configuration options (like enabling L/R check or subpixel) can be added in this box too
- Improved handling of depth frames (to be the same as is [here](https://gist.github.com/alex-luxonis/b8b3f900ea27e42ded37c70eeb814bb7#file-sub_pixel-py-L98))
- Fixed bug with incorrect handling of `-s / --show` parameter that displayed more frames than was requested by user
- Added mouse click tracker that allows to click on depth / disparity image and see the value at a specific location
- Introduced no-nn mode that allows to disable nn inference and adjusted the created camera nodes by more strictly checking when they're needed so that now `-sub / --subpixel` mode can be accessed with the demo

Demo: (click)
[demo](http://www.youtube.com/watch?v=7o6jrLYkd5Q "demo")

### DIFF
--- a/depthai_demo.py
+++ b/depthai_demo.py
@@ -9,7 +9,7 @@ from depthai_helpers.version_check import check_depthai_version
 import platform
 
 from depthai_helpers.arg_manager import parse_args
-from depthai_helpers.config_manager import BlobManager, ConfigManager
+from depthai_helpers.config_manager import ConfigManager
 from depthai_helpers.utils import frame_norm, to_planar, to_tensor_result, load_module
 
 print('Using depthai module from: ', dai.__file__)
@@ -132,11 +132,11 @@ with dai.Device(pm.p.getOpenVINOVersion(), device_info, usb2Mode=conf.args.usb_s
     if conf.useCamera or conf.args.sync:
         pv = PreviewManager(fps, display=conf.args.show, colorMap=conf.getColorMap(), disp_multiplier=disp_multiplier)
 
-        if conf.args.camera == "left" or conf.useDepth:
+        if conf.leftCameraEnabled:
             pm.create_left_cam(mono_res, conf.args.mono_fps)
-        if conf.args.camera == "right" or conf.useDepth:
+        if conf.rightCameraEnabled:
             pm.create_right_cam(mono_res, conf.args.mono_fps)
-        if conf.args.camera == "color":
+        if conf.rgbCameraEnabled:
             pm.create_color_cam(nn_manager.input_size if conf.useNN else (300, 300), rgb_res, conf.args.rgb_fps, conf.args.full_fov_nn, conf.useHQ)
 
         if conf.useDepth:
--- a/depthai_demo.py
+++ b/depthai_demo.py
@@ -180,8 +180,9 @@ with dai.Device(pm.p.getOpenVINOVersion(), device_info, usb2Mode=conf.args.usb_s
             if queue_name in [Previews.disparity_color.name, Previews.disparity.name, Previews.depth.name]:
                 Trackbars.create_trackbar('Disparity confidence', queue_name, 0, 255, conf.args.disparity_confidence_threshold,
                          lambda value: pm.update_depth_config(device, dct=value))
-                Trackbars.create_trackbar('Bilateral sigma', queue_name, 0, 250, conf.args.sigma,
-                         lambda value: pm.update_depth_config(device, sigma=value))
+                if queue_name == Previews.depth.name:
+                    Trackbars.create_trackbar('Bilateral sigma', queue_name, 0, 250, conf.args.sigma,
+                             lambda value: pm.update_depth_config(device, sigma=value))
                 if conf.args.stereo_lr_check:
                     Trackbars.create_trackbar('LR-check threshold', queue_name, 0, 10, conf.args.lrc_threshold,
                              lambda value: pm.update_depth_config(device, lrc_threshold=value))
--- a/depthai_demo.py
+++ b/depthai_demo.py
@@ -177,10 +177,10 @@ with dai.Device(pm.p.getOpenVINOVersion(), device_info, usb2Mode=conf.args.usb_s
 
     if conf.useCamera:
         def create_queue_callback(queue_name):
-            if queue_name in [Previews.disparity_color.name, Previews.disparity.name, Previews.depth.name]:
+            if queue_name in [Previews.disparity_color.name, Previews.disparity.name, Previews.depth.name, Previews.depth_raw.name]:
                 Trackbars.create_trackbar('Disparity confidence', queue_name, 0, 255, conf.args.disparity_confidence_threshold,
                          lambda value: pm.update_depth_config(device, dct=value))
-                if queue_name == Previews.depth.name:
+                if queue_name in [Previews.depth_raw.name, Previews.depth.name]:
                     Trackbars.create_trackbar('Bilateral sigma', queue_name, 0, 250, conf.args.sigma,
                              lambda value: pm.update_depth_config(device, sigma=value))
                 if conf.args.stereo_lr_check:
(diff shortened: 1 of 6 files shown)


themarpe (approved):
LGTM

depthai_helpers/managers.py
```diff
@@ -29,6 +23,21 @@ def convert_disparity_frame(packet, manager):
 def convert_disparity_to_color(disparity, manager):
     return cv2.applyColorMap(disparity, manager.colorMap)
 
+
+def convert_depth_frame(packet, manager):
+    depth_frame = packet.getFrame()
+    dispScaleFactor = getattr(manager, "dispScaleFactor", None)
+    if dispScaleFactor is None:
+        baseline = 75  # mm
```
szabi-luxonis, line 31: Baseline and focal length should be obtained from EEPROM data, if available, otherwise defaulted to these values.
This is not going to work well on smaller baseline boards.

szabi-luxonis (changes requested):

depthai_helpers/managers.py
```diff
@@ -37,52 +115,130 @@ class Previews(enum.Enum):
...
         self.disp_multiplier = disp_multiplier
-
-    def create_queues(self, device):
-        def get_output_queue(name):
+        self.mouse_tracker = MouseClickTracker() if mouseTracker else None
+
+    def create_queues(self, device, callback=lambda *a, **k: None):
+        calib = device.readCalibration()
+        eeprom = calib.getEepromData()
+        cam_info = eeprom.cameraData[calib.getStereoLeftCameraId()]
```
szabi-luxonis, line 168: This will fail on OAK1, or if calibration is empty/corrupted.

szabi-luxonis: From previous review comments:
```
Baseline and focal length should be obtained from EEPROM data, if available, otherwise defaulted to these values.
This is not going to work well on smaller baseline boards.
```

szabi-luxonis (changes requested):

depthai_helpers/managers.py
VanDavv: Ok, updated so that OAK-1 works without issues with this script too.
> Baseline and focal length should be obtained from EEPROM data, if available, otherwise defaulted to these values.
> This is not going to work well on smaller baseline boards.
when the calibration is empty,  `readCalibration` or `getEepromData` would throw an error or return null @szabi-luxonis?

depthai_demo.py
```diff
@@ -112,46 +130,69 @@ def print_sys_info(info):
...
-        if conf.args.camera == "left" or conf.useDepth:
+        if conf.leftCameraEnabled:
             pm.create_left_cam(mono_res, conf.args.mono_fps)
-        if conf.args.camera == "right" or conf.useDepth:
+        if conf.rightCameraEnabled:
             pm.create_right_cam(mono_res, conf.args.mono_fps)
-        if conf.args.camera == "color":
-            pm.create_color_cam(rgb_res, conf.args.rgb_fps, conf.args.full_fov_nn, conf.useHQ)
+        if conf.rgbCameraEnabled:
+            pm.create_color_cam(nn_manager.input_size if conf.useNN else (300, 300), rgb_res, conf.args.rgb_fps, conf.args.full_fov_nn, conf.useHQ)
```
szabi-luxonis, line 140: for these king of hardcoded constansts would be better using a global const, so it's more transparent, easier to change, not hidden in the middle of the code somewhere.
300x300 is default in case NN doesn't specify input size? Or it's if NN was not requested? If latter then it would be better setting it to the same aspect ratio as the camera, e.g. 576x324, so it covers the full FOV

depthai_demo.py
```diff
@@ -112,46 +130,69 @@ def print_sys_info(info):
...
     if len(conf.args.report) > 0:
         pm.create_system_logger()
 
-    nn_pipeline = nn_manager.create_nn_pipeline(pm.p, pm.nodes, shaves=conf.args.shaves, use_depth=conf.useDepth,
-                                                use_sbb=conf.args.spatial_bounding_box and conf.useDepth,
-                                                minDepth=conf.args.min_depth, maxDepth=conf.args.max_depth,
-                                                sbbScaleFactor=conf.args.sbb_scale_factor)
-    pm.create_nn(nn=nn_pipeline, sync=conf.args.sync)
+    if conf.useNN:
+        nn_pipeline = nn_manager.create_nn_pipeline(pm.p, pm.nodes, shaves=conf.args.shaves, use_depth=conf.useDepth,
```
szabi-luxonis, line 163: try to be more consistent with naming across the codebase, either camelCase (as core) or snake_case

depthai_demo.py
```diff
@@ -112,46 +130,69 @@ def print_sys_info(info):
...
+            break
+
     if conf.useCamera:
-        pv.create_queues(device)
+        def create_queue_callback(queue_name):
+            if queue_name in [Previews.disparity_color.name, Previews.disparity.name, Previews.depth.name, Previews.depth_raw.name]:
+                Trackbars.create_trackbar('Disparity confidence', queue_name, 0, 255, conf.args.disparity_confidence_threshold,
+                         lambda value: pm.update_depth_config(device, dct=value))
+                if queue_name in [Previews.depth_raw.name, Previews.depth.name]:
+                    Trackbars.create_trackbar('Bilateral sigma', queue_name, 0, 250, conf.args.sigma,
```
szabi-luxonis, line 190: Not sure, is it possible to dynamically change the maximum value on trackbar?
It would be fun to play with extreme values. 65535 is the max for bilateral sigma, 127/8 for LR-check threshold.
Nice to have if it's easy.

depthai_demo.py
```diff
@@ -172,51 +213,74 @@ def print_sys_info(info):
...
-                frame_nn.setWidth(nn_manager.input_size[0])
-                frame_nn.setHeight(nn_manager.input_size[1])
-                frame_nn.setData(to_planar(scaled_frame))
-                nn_in.send(frame_nn)
-                seq_num += 1
+                if nn_in is not None:
+                    scaled_frame = cv2.resize(host_frame, nn_manager.input_size)
+                    frame_nn = dai.ImgFrame()
+                    frame_nn.setSequenceNum(seq_num)
+                    frame_nn.setType(dai.RawImgFrame.Type.BGR888p)
```
szabi-luxonis, line 240: TODO handle input frame type when we will have blob parsing.
Also use `ImgFrame.Type.BGR888p` instead `RawImgFrame.Type.BGR888p`

szabi-luxonis:
Otherwise LGTM for this PR

szabi-luxonis (approved):

szabi-luxonis:
Update to latest develop before merge.

depthai_demo.py
VanDavv: It's when the nn is disabled and no input size was specified. I'll adjust it and store as a variable / param

depthai_demo.py
VanDavv: @szabi-luxonis I didn't want to expand unnecessarily the cli args, so I made them configurable by env variables

```
DISP_CONF_MIN = int(os.getenv("DISP_CONF_MIN", 0))
DISP_CONF_MAX = int(os.getenv("DISP_CONF_MAX", 255))
SIGMA_MIN = int(os.getenv("SIGMA_MIN", 0))
SIGMA_MAX = int(os.getenv("SIGMA_MAX", 250))
LRCT_MIN = int(os.getenv("LRCT_MIN", 0))
LRCT_MAX = int(os.getenv("LRCT_MAX", 10))
```

szabi-luxonis: By dynamic I meant if the user sets the slider to max value, increase the max value on the slider itself (or better shift the min max range) , so it can be set to higher than initial

PR #415, opened 2021-07-01
Added min allowed SquareSize

saching13:
Set min allowed square size for calibration to be greater than 2.2 cm.


Luxonis-Brandon:
Is it the case that below 2.2cm that the calibration is not sufficiently accurate?

saching13:
> Is it the case that below 2.2cm that the calibration is not sufficiently accurate?

It looked like it. User in Discord had good depth image but values were off. Calibrating by increasing square size from 1.8cm to 2.4+ approx fixed issue for him.

Luxonis-Brandon:
Got it, thanks.  We should add that to the calibration documentation as well then @saching13 so that folks know the `why`.

And thanks for the catch/debug with the Discord user!

VanDavv (approved):
LGTM, thanks!

PR #422, opened 2021-07-13
Enable full FOV output from RGB camera

VanDavv:
With this PR, the preview from the RGB camera is adjusted to meet the full FOV aspect ratio

image

### DIFF
--- a/depthai_demo.py
+++ b/depthai_demo.py
@@ -123,7 +123,7 @@ if conf.useNN:
         model_name=conf.getModelName(),
         model_dir=conf.getModelDir(),
         source=conf.getModelSource(),
-        full_fov=conf.args.full_fov_nn or not conf.useCamera,
+        full_fov=not conf.args.disable_full_fov_nn or not conf.useCamera,
         flip_detection=conf.getModelSource() in ("rectified_left", "rectified_right") and not conf.args.stereo_lr_check
     )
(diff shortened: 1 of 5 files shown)


Luxonis-Brandon:
Yes this is a lot nicer!  

image

So a quick question - the 1920x1080 (I presume that's the sensor config) is cropped out of the aspect ratio of the full image sensor, is that right?

Luxonis-Brandon:
But it looks like the spatial location calculator is being used wrong.  Look at the location of the TV monitor in the background.  

I think the wrong locations are being used between depth and RGB.

Luxonis-Brandon:
Here's a video showing it:

https://photos.app.goo.gl/RyohNDddfpEx6MF59

VanDavv:
Yes, I saw that too, and I think that's due to the lack of RGB-Depth alignment. Started to work on it [here](https://github.com/luxonis/depthai/tree/depth_rgb_alignment) but had come issues

Luxonis-Brandon:
So I think it's different than RGB depth alignment.  As w/out this PR the distances seem correct.

Luxonis-Brandon:
Here's a video of what's in `main` for example:
https://photos.app.goo.gl/nayGAPHxqQWvYzU56

VanDavv:
Full FOV preview was enabled in #433 due to performance requirements

PR #434, opened 2021-08-02
Add demo tweaks

VanDavv:
This PR addresses improvements ideas from #433.
With this PR:
- RGB preview image is now scaled before any overlay is applied (reported [here](https://github.com/luxonis/depthai/pull/433#issuecomment-890573026))
- Applied text printing improvements (reported [here](https://github.com/luxonis/depthai/pull/433#issuecomment-890580600))
  image
- If low bandwidth mode is detected (PoE usecase), it will use `disparity_color` stream by default instead of `depth`, because the latter one is not yet encodable with MJPEG

### DIFF
--- a/depthai_demo.py
+++ b/depthai_demo.py
@@ -278,9 +278,8 @@ with dai.Device(pm.p.getOpenVINOVersion(), device_info, usb2Mode=conf.args.usb_s
                             "fontScale": 0.4,
                             "thickness": 1
                         }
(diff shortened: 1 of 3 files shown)


Luxonis-Brandon:
Looks great to me.  Will test today.

Luxonis-Brandon (approved):
Ignore the terrible-looking depth; I am using an uncalibrated pre-production OAK-D-POE (need to calibrate it soon).  

image

This looks good to me. 

LGTM

PR #435, opened 2021-08-02
Fix OpenPose nn

VanDavv:
This PR fixes the` openpose2` neural network, it crashed on main with

```
  File "./depthai_demo.py", line 264, in <module>
    nn_data = nn_manager.decode(in_nn)
  File "/Users/leeroy/depthai/depthai_helpers/managers.py", line 469, in decode
    return self.handler.decode(self, in_nn)
  File "/Users/leeroy/depthai/resources/nn/openpose2/handler.py", line 152, in decode
    valid_pairs, invalid_pairs = getValidPairs(outputs, w, h, detected_keypoints)
  File "/Users/leeroy/depthai/resources/nn/openpose2/handler.py", line 46, in getValidPairs
    pafA = outputs[0, mapIdx[k][0], :, :]
IndexError: index 47 is out of bounds for axis 1 with size 46
```

The source of this issue was the `[::-1]` transformation of tensor dimensions. The bug that it resolved was fixed in the latest API release and after the update, the transformation started to bring the opposite effect (yielded incorrect tensor dims)

Also, since we're using full FOV now for preview, had to adjust `handler.py` for openpose to scale the NN results accordingly

### DIFF
--- a/depthai_helpers/utils.py
+++ b/depthai_helpers/utils.py
@@ -26,11 +26,11 @@ def to_tensor_result(packet):
     data = {}
     for tensor in packet.getRaw().tensors:
         if tensor.dataType == dai.TensorInfo.DataType.INT:
-            data[tensor.name] = np.array(packet.getLayerInt32(tensor.name)).reshape(tensor.dims[::-1])
+            data[tensor.name] = np.array(packet.getLayerInt32(tensor.name)).reshape(tensor.dims)
         elif tensor.dataType == dai.TensorInfo.DataType.FP16:
-            data[tensor.name] = np.array(packet.getLayerFp16(tensor.name)).reshape(tensor.dims[::-1])
+            data[tensor.name] = np.array(packet.getLayerFp16(tensor.name)).reshape(tensor.dims)
         elif tensor.dataType == dai.TensorInfo.DataType.I8:
-            data[tensor.name] = np.array(packet.getLayerUInt8(tensor.name)).reshape(tensor.dims[::-1])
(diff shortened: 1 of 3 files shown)


VanDavv:
One note here: since we're using a full FOV preview, the results are being scaled to fit the target frame. This is mostly fine except that networks like deeplab (that return a full frame overlay) tend not to scale well and can be inconsistent around the edges. The best way to visualize them is to use `-s nn_input` as then the NN results will be applied without any scaling

themarpe (approved):
LGTM - might be good to check any other places for the invalid transformation, mainly depthai-experiments

Luxonis-Brandon:
Yes the handling of the NN data looks wrongly-scaled @VanDavv :

image

So I think we should do a follow-up PR to either fix these directly, or if we can't, make the preview size for such networks be the the same aspect ratio of the neural network.

Thoughts?

Luxonis-Brandon:
We should merge this first though... as this is better than crashing.  But we should then circle back to make sure the overlays are correct - doing whatever is necessary to ensure that.

Luxonis-Brandon (approved):
I'd say let's merge.  Please circle back to fix the [scaling error](https://github.com/luxonis/depthai/pull/435#issuecomment-891394387) though with a new PR.

VanDavv:
Sure, will work on this and circle back with PR

VanDavv:
@Luxonis-Brandon the PR with fixes is [here](https://github.com/luxonis/depthai/pull/438)

Luxonis-Brandon:
Thanks!

PR #438, opened 2021-08-07
Fix OpenPose & Deeplab nn mapping

VanDavv:
This PR fixes the nn overlay issue reported in #435 .
Due to this, both deeplabv3p_person and openpose2 were displaying it's results on larger frames with scaling issue (they were not showing the results correctly)

Now, the openpose2 shows correct output on the color preview (the results are off on the depth preview due to no rgb-depth alignment)
frame_009_delay-0 04s

Also, the deeplab network is now correctly presenting it's results too
Screenshot from 2021-08-07 12-16-19


Erol444 (approved):
Works great for me as well, thanks:)

PR #440, opened 2021-08-09
Revert "Increase sigma range and default value"

szabi-luxonis:
Reverts luxonis/depthai#416


Luxonis-Brandon (approved):
Looks good to me.  Works on USB and POE alike and fixes the stall issue seen on USB on Mac.

PR #441, opened 2021-08-10
Fix mouse tracker on disparity stream over PoE

VanDavv:
This PR fixes the` disparity_color` mouse tracker display. With PoE devices, it showed `None` instead of the correct pixel value.
Main issue here was that the mouse tracker mechanism took values from MJPEG encoded frame, but I also cleaned up the code related to mouse tracking a bit.
image


Erol444 (approved):
LGTM!

szabi-luxonis (approved):

PR #444, opened 2021-08-10
Calculate optimal number of shaves based on cli params

VanDavv:
This PR introduces shaves calculation method based on current cli params


szabi-luxonis (approved):
LGTM

PR #447, opened 2021-08-11
Demo fixes

### LINKED ISSUE #454: [BUG] Previews does not have a property called host depthai_demp.py
**Describe the bug**

image

**To Reproduce**

1. Update the code to the latest main branch and run 
2. `python depthai_demo.py -cnn road-segmentation-adas-0001 -sync -vid /media/Workspace/Learning/Github/depthai/videos/CamVid.mp4 -sh 8`

**Expected behavior**
It will run road segmentation successfully.

**Additional context**
After I looked into the code, it seems that the latest code does not define `host` property on [Previews class](https://github.com/luxonis/depthai/blob/f8356a7ad7a09b6c51da6b3ea4d2744b4272ec44/depthai_helpers/managers.py#L166), but somehow the the [code in depthai_demo.py](https://github.com/luxonis/depthai/blob/f8356a7ad7a09b6c51da6b3ea4d2744b4272ec44/depthai_demo.py#L209) is still trying to use it.

VanDavv:
This PR introduces various demo fixes:
- Fixed `--sync` mode
- Fixed default scaling of the preview window
- Fixed `nn_input` preview if the source is set to rectified left/right
- Fixed mouse tracker on grayscale frames
- Fixed NN results scaling for wide-input networks (aspect ratio > 1)
- Fixed host nn input (resolves #454 )


Luxonis-Brandon:
The bounding box mapping calculations are wrong in at least this version:

image
image


@themarpe or @Erol444 we'll need you to review as we're at the point of "circular" iteration on this bounding box work rather than spiral iteration.  (As in we're not getting closer to the right answer over iterations, we're just circling the right answer with progressively wrong-in-different-way wrong answers.)

Luxonis-Brandon:
Here's a video showing it:
https://photos.app.goo.gl/3duetWx4Y76zM8wx6

VanDavv:
@Luxonis-Brandon I've enabled full fov, as it's way easier to map the results with full fov (as we don't crop the input frame then) - so "by default" users will get better nn results mapping. 

I'm still thinking about how to do the correct nn mapping for left/right cameras (640x400) when we use the color camera as a source and don't use full fov (so the cropping occurs). It gets even harder with wide-input networks, like `face-detection-adas-0001` (672x384)
image

I will continue to investigate this (and brainstorm with @szabi-luxonis / @themarpe / @Erol444 the best solution here) , but since this issue is also present on `main`, I'd merge this PR for other fixes

Luxonis-Brandon:
Agreed @VanDavv - thanks.  Given that it is in main, I think it's the right call to merge the fixes here.

Erol444 (approved):
LGTM, thanks:)

PR #451, opened 2021-08-12
Add an error and solution link when detected M1 chip

VanDavv:
This PR introduces a better handling for M1 chips in `install_requirements.py` file. Will throw an error, as we don't have prebuilt binaries anyways, but will also provide a link to discuss thread with a solution

image

I'll move this check together with #448 into depthai-python installer later


Erol444 (approved):
Thanks:)

PR #458, opened 2021-08-16
Fix full FOV option for mono outputs

VanDavv:


Luxonis-Brandon:
Thanks for making this!

Erol444 (approved):

PR #459, opened 2021-08-19
Fix FPS calculation on host

VanDavv:
This PR improves FPS calculations - as suggested by @alex-luxonis, `deque` was used and the average FPS is now more reliable.

Running the same command (`python3 depthai_demo.py`) in this PR printed

```
=== TOTAL FPS ===
[color]: 35.3
[depth_raw]: 33.1
[depth]: 33.1
[nn]: 32.3
```

Whereas on `main`

```
=== TOTAL FPS ===
[color]: 27.9
[depth_raw]: 27.6
[depth]: 27.6
[nn]: 27.9
```

### DIFF
--- a/depthai_helpers/managers.py
+++ b/depthai_helpers/managers.py
@@ -1,3 +1,4 @@
+import collections
 import json
 import math
 import sys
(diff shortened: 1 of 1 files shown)


Erol444 (approved):

depthai_helpers/managers.py
```diff
@@ -551,46 +552,43 @@ class FPSHandler:
...
+        self.iter_cnt += 1
 
     def tick(self, name):
-        if name in self.ticks:
-            self.ticks_cnt[name] += 1
-        else:
-            self.ticks[name] = time.monotonic()
-            self.ticks_cnt[name] = 0
+        if name not in self.ticks:
+            self.ticks[name] = collections.deque(maxlen=10)
```
alex-luxonis, line 577: The default queue length of 10 is a bit small, and doesn't filter well the jitter. I'd suggest about 30, or maybe even better like 100 (a max window of 3.33s at 30 fps).
Ideally the averaging window length in time would be runtime-adjusted based on the FPS, still having a max-sized `deque`, but taking only the last `N` samples in the FPS calc, with 
`N = math.ceil(FPS / predefined_window_length_seconds) + 1`.
But I think it's fine to just increase the max window length for now.

themarpe:
What about just counting number of frames that arrive in one second, and then resetting the counter with next second? IMO works well

depthai_helpers/managers.py
```diff
@@ -551,46 +552,43 @@ class FPSHandler:
...
+        self.ticks[name].append(time.monotonic())
 
     def tick_fps(self, name):
-        if name in self.ticks:
-            time_diff = time.monotonic() - self.ticks[name]
-            return self.ticks_cnt[name] / time_diff if time_diff != 0 else 0
+        if name in self.ticks and len(self.ticks[name]) > 1:
+            time_diff = self.ticks[name][-1] - self.ticks[name][0]
+            return len(self.ticks[name]) / time_diff if time_diff != 0 else 0
```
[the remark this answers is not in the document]
VanDavv: thanks!

depthai_helpers/managers.py
VanDavv: Thanks! yes, agreed, I think 100 will suit well for now, committed already

VanDavv:
> What about just counting number of frames that arrive in one second, and then resetting the counter with next second? IMO works well

@themarpe I think that the use of `deque` could help filter out the jitter and show the correct number of FPS e.g. when NN would have 1.5 FPS -`deque` would show it correctly, whereas the simpler approach would vary between 1 and 2 fps

alex-luxonis:
It's possible to have a good floating point value even as @themarpe suggested, by calculating when the given time was exceeded, with the precise elapsed interval, like:
```
        tnow = time.monotonic()
        tdiff = tnow - self.prevTime[name]
        if tdiff >= 1:
            self.fps[name] = self.count[name] / tdiff
            self.count[name] = 0
            self.prevTime[name] = tnow
```

The only advantage of storing every frame's timestamp is we can provide an updated FPS average every frame (from the second frame on). Whether this is good or not (i.e. too hard to read due to frequent updates), is debatable.

Also, counting per second has a slight disadvantage it shows 0 FPS until 1 second elapsed.

alex-luxonis (approved):

Issue #455, opened 2021-08-16
[BUG] Video window doesn't show combined/addWeight image with mask


franva:
**Check if issue already exists**
![image](https://user-images.githubusercontent.com/5151146/129508358-3646a108-589a-4d78-82f5-28f2cc33a21c.png)

**Describe the bug**
The `handler.py` is actually working means it does return correct segmentation data, BUT the host window does not show the mask/overlay on top of the original image.

**To Reproduce**
`handler.py`

```
from typing import overload
from depthai_helpers.managers import Previews
from resources.nn.colormap import ColorMap
import cv2
import numpy as np

from depthai_helpers.utils import to_tensor_result


def decode(nn_manager, packet):
    # [print(f"Layer name: {l.name}, Type: {l.dataType}, Dimensions: {l.dims}") for l in packet.getAllLayers()]
    # after squeeze the data.shape is 4,512, 896
    data = np.squeeze(to_tensor_result(packet)["L0317_ReWeight_SoftMax"])
    class_colors = ColorMap.COLORS
    class_colors = np.asarray(class_colors, dtype=np.uint8)

    indices = np.argmax(data, axis=0)
    indices += 8
    output_colors = np.take(class_colors, indices, axis=0)
    return output_colors



def draw(nn_manager, data, frames):
    if len(data) == 0:
        return
    for name, frame in frames:
        if name in (nn_manager.source, 'host'):
            resized = frame
            if frame.shape != data.shape:
                # resize the data to fit frame
                newsize = frame.shape[:2][1],frame.shape[:2][0]
                resized = cv2.resize(data, newsize)

            # overlayed = cv2.addWeighted(resized, 0.5, data, 0.5, 0)
            overlayed = (frame * 0.5 + resized * 0.5).astype('uint8')

            return overlayed
            # return data

# def draw(nn_manager, data, frames):
#     if len(data) == 0:
#         return
#     for name, frame in frames:
#         if name in (Previews.color.name, Previews.nn_input.name, 'host'):
#             scale_factor = frame.shape[0] / nn_manager.input_size[1]
#             resize_w = int(nn_manager.input_size[0] * scale_factor)
#             resized = cv2.resize(data, (resize_w, frame.shape[0])).astype(data.dtype)
#             offset_w = int(frame.shape[1] - nn_manager.input_size[0] * scale_factor) // 2
#             tail_w = frame.shape[1] - offset_w - resize_w
[...]

PR #460, opened 2021-08-19
Add road-segmentation-adas-0001 handler

VanDavv:
This PR adds a `road-segmentation-adas-0001` handler to depthai-demo. Thanks to @franva for his pointers in #455 that made this PR easier to implement.

### Demo

ezgif com-gif-maker (1)

### To reproduce

```
depthai_demo.py -cnn road-segmentation-adas-0001 -vid https://www.youtube.com/watch?v=lCmuKhHVAIg --sync
```

### TODO

- [x] Resolve why the detection overlay is sometimes applied twice

### DIFF
--- a/depthai_demo.py
+++ b/depthai_demo.py
@@ -263,7 +263,6 @@ with dai.Device(pm.p.getOpenVINOVersion(), device_info, usb2Mode=conf.args.usb_s
 
                 if not conf.args.sync:
                     host_frame = raw_host_frame
-
                 fps.tick('host')
 
             if nn_out is not None:
@@ -290,10 +289,12 @@ with dai.Device(pm.p.getOpenVINOVersion(), device_info, usb2Mode=conf.args.usb_s
                         return return_frame if return_frame is not None else frame
                 pv.show_frames(callback=show_frames_callback)
             elif host_frame is not None:
+                debug_host_frame = host_frame.copy()
+                cv2.imshow("TEST", debug_host_frame)
                 if conf.useNN:
-                    nn_manager.draw(host_frame, nn_data)
-                fps.draw_fps(host_frame, "host")
-                cv2.imshow("host", host_frame)
+                    nn_manager.draw(debug_host_frame, nn_data)
+                fps.draw_fps(debug_host_frame, "host")
+                cv2.imshow("host", debug_host_frame)
 
             if log_out:
                 logs = log_out.tryGetAll()
--- a/depthai_demo.py
+++ b/depthai_demo.py
@@ -290,7 +290,6 @@ with dai.Device(pm.p.getOpenVINOVersion(), device_info, usb2Mode=conf.args.usb_s
                 pv.show_frames(callback=show_frames_callback)
             elif host_frame is not None:
                 debug_host_frame = host_frame.copy()
-                cv2.imshow("TEST", debug_host_frame)
                 if conf.useNN:
                     nn_manager.draw(debug_host_frame, nn_data)
                 fps.draw_fps(debug_host_frame, "host")
--- a/resources/nn/road-segmentation-adas-0001/handler.py
+++ b/resources/nn/road-segmentation-adas-0001/handler.py
@@ -1,0 +1,33 @@
+import cv2
+import numpy as np
+
+from depthai_helpers.managers import Previews
+from depthai_helpers.utils import to_tensor_result
+
+
+def decode(nn_manager, packet):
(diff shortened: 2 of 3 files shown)


Erol444 (approved):
Just tried it out, works great, thanks!

franva:
tried out, it looks great~! love the team work~!!! thanks @VanDavv 

Could you please explain 
1. Why doesn't my code draw the mask on top of the video?
2. Why do you only do the "proper resize" for color source, but not for the "host"?

For question 1, I guess the answer hides on the line that you use `np.hstack()` ? But I don't know why exactly it works in this way, but not in my old code.  Appreciate if you could explain it in details.

Erol444:
@franva  have you tried running `depthai_demo.py -cnn road-segmentation-adas-0001 -vid https://www.youtube.com/watch?v=lCmuKhHVAIg --sync`? For me, it draws the mask on top of the video as expected (like on the screenshot above). Have you pulled the latest commit?

franva:
@Erol444  sorry I didn't specify it enough.

Yes, I pulled the latest commit.

When I referred to the "my code", I mean the code I posted here : https://github.com/luxonis/depthai/issues/455

I tried to stack the original image with the mask like:
`overlayed = cv2.addWeighted(resized, 0.5, data, 0.5, 0)`
or 
`overlayed = (frame * 0.5 + resized * 0.5).astype('uint8')`

But none of them worked.

VanDavv:
Hi @franva, sure I can give more background on the implementation:

> Why doesn't my code draw the mask on top of the video?

The `draw` method receives frames that are later displayed in the preview window. The thing is that this method does not return anything, and to apply changes to the frame you have to modify the frame itself. OpenCV functions for drawing - `drawX` - modify received frame instead of creating a new one, and I wanted to follow the same convention. In your code, you created a new frame called `overlayed` and then returned it - but in fact, the return value was omitted and not consumed by the calling code. I modified your `addWeighted` call to actually save its result to the `frame` object (6th argument - `dst` - specifies the result array)

```diff
- overlayed = cv2.addWeighted(resized, 0.5, data, 0.5, 0)
+ cv2.addWeighted(frame, 0.5, data, 0.5, 0, frame)
```

This way I actually modify the received frame and results are visible in preview window

> Why do you only do the "proper resize" for color source, but not for the "host"?

We have 3 different cases here to handle
- Source is from host (e.g. video file)
- Source is from RGB camera with Full FOV (default)
- Source is from RGB camera without Full FOV (`-dff` flag)

I also assume that we want only to display the results onto RGB frames (without Mono/Depth) as it's most accurate and easiest to implement (we can extend support for them later)
With this assumption, we have actually only 3 frame types possible - `color`, `nn_input` and `host`.

Now, with `nn_input`, we can go ahead and map the nn results directly as the frame size will be the same as the output frame. With `host`, it's also pretty easy, since we pass the nn a scaled video frame, so it's enough to scale the nn result back to video frame shape and we'll have correctly aligned results.

For the `color` frame, things get a bit more complicated. If the demo is called with `-dff / --disable_full_fov` flag, the image from the RGB camera is first cropped to meet nn aspect ratio (center crop) and then is scaled. So in order to correctly map the results back to full fov, we have to calculate the width of the crop and then correctly apply the results with offset.
To give some in-code references:
- `resize_w` represents the results frame resize width (resized frame - `resized`)
- `offset_w` / `tail_w` represents crop width, from left and right respectively
- `stacked` is the result frame, where I added the previously cropped regions.

Thankfully, the nn input size is 896x512, so the actual total crop width is `30px` for 1080p - so the `offset_w` and `tail_w` variables should be around `15px` respectively, which is pretty much unnoticeable - as we also scale the `color` preview to 1/3 of it's original size to make the preview look better.

franva:
@VanDavv  thank you very much for such a detailed explanation!

For Q1, I now know why it didn't work when I was trying my old code. There are 2 reasons:
1. I didn't notice that the returned frame is totally omitted
2. even when I started with the `cv2.addWeighted(resized, 0.5, data, 0.5, 0, frame)` I could not see the overlay mask. I just noticed the reason today, it's because of the colors I used e.g.
```
SKY=[28,51,71]
BUILDING=[28,28,28]
POLE=[60,60,60]
etc....
```
They are somehow almost invisible and after I changed the colors to the "pure colors" e.g. `[0, 0, 0], [0, 255, 0], [255, 0, 0], [0, 0, 255]`, I started to see the mask.

For Q2, thanks for the detailed explanation.  The detailed explanation can be copied and pasted into the documentation for other users who have the same question :)

PR #465, opened 2021-08-24
Handle full FOV in openpose2 handler

VanDavv:
This PR adds a better handling for full FOV when running `openpose2` network


Erol444 (approved):
works as expected now, thanks!

PR #469, opened 2021-08-30
Introduce DepthAI-SDK package

VanDavv:
DepthAI SDK is a Python package, containing convenience classes and functions that help in most common tasks while using DepthAI API.

Package preview - https://test.pypi.org/project/depthai-sdk/

To install package right now, please run the following command

```
pip install --extra-index-url https://test.pypi.org/simple/ depthai-sdk
```

To see this package in action, you can check the migrated experiments here - https://github.com/luxonis/depthai-experiments/pull/192
- `gen2-people-counter`: https://github.com/luxonis/depthai-experiments/tree/sdk/gen2-people-counter
- `gen2-human-pose`: https://github.com/luxonis/depthai-experiments/tree/sdk/gen2-human-pose
- `gen2-road-segmentation`: https://github.com/luxonis/depthai-experiments/tree/sdk/gen2-road-segmentation

TODO
- [x] Add generating docs from python docstrings using Sphinx - https://github.com/luxonis/depthai/pull/469/commits/e1f2fab1fd78a1bbddce7884550d3f3f37f52434
- [x] Add docs subproject (on docs.luxonis.com) - https://docs.luxonis.com/projects/sdk/en/latest/
- [x] Add python docstrings - https://docs.luxonis.com/projects/sdk/en/latest/api/
- [x] Add Examples/Recipies/Tutorials that will show the SDK usage (basic usecases for now) - https://docs.luxonis.com/projects/sdk/en/latest/getting_started


Erol444 (approved):
LGTM, I saw that there were other really useful improvements as well - like running from video won't decrease/resize the original video, but resize the BBs to fit original high-quality video.
Also should we move the handlers.py to the SDK? That would be really useful IMO. And if we do that, we should move supported models to the docs (from depthai-website docs).

VanDavv:
Thanks @Erol444. Yes, we should add them for sure. Eventually we would have a pair of `decode` and `show` functions for each NN we support, under a convenient module. So something like below shows the "usage" idea

```
from depthai_sdk.nn.openpose2 import decode, show
```

This PR will start a `1.0.0` version and we'll add improvements like these in the following PRs, bumping the versions and making a new release

Erol444:
Awesome, thanks!

depthai_sdk/src/depthai_sdk/__init__.py
```diff
@@ -0,0 +1,4 @@
+from .fps import *
+from .previews import *
+from .utils import *
+from .managers import *
```
themarpe, line 4: Does this remove the 'namespaces' when doing `import depthai_sdk` so `depthai_sdk.managers.PipelineManager` is then `depthai_sdk.PipelineManager`?

Just sane checking if this is a good idea

depthai_sdk/docs/Makefile
```diff
@@ -0,0 +1,20 @@
+# Minimal makefile for Sphinx documentation
```
themarpe, line 1: Afaik not needed

depthai_sdk/docs/make.bat
```diff
@@ -0,0 +1,35 @@
+@ECHO OFF
```
themarpe, line 1: Afaik needed even less :p

depthai_sdk/README.md
```diff
@@ -0,0 +1,27 @@
...
+To install this package, run the following command in your terminal window
+
+```
+$ pip install depthai-sdk
+```
+
+> :warning: **If you're using Raspberry Pi**, providing a Pi Wheels extra package url can significantly speed up the instalation process by providing prebuilt binaries for OpenCV
+> ```
+> $ pip install --extra-index-url https://www.piwheels.org/simple/ depthai-sdk
+> ```
```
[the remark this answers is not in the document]
VanDavv: Yes, I want to make 1.0.0 release, that's why in readme I included the "after release" text that will end up in main PyPi

And regarding the warning - I think that it's something that may not help but won't hurt either, would leave for now

depthai_demo.py
```diff
@@ -172,35 +180,41 @@ def fn(value):
...
+        )
 
     # Start pipeline
-    device.startPipeline(pm.p)
+    device.startPipeline(pm.pipeline)
     pm.create_default_queues(device)
-    nn_in = device.getInputQueue(nn_manager.input_name, maxSize=1, blocking=False) if not conf.useCamera and conf.useNN else None
-    nn_out = device.getOutputQueue(nn_manager.output_name, maxSize=1, blocking=False) if conf.useNN else None
+    if conf.useNN:
+        nn_manager.createQueues(device)
```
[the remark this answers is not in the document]
VanDavv: Yes, and the `camelCase` I think is a go here - it's not following python standards but combines well with our API. Didn't want to include it in this release though as the size of it would be too big, and it's something we can do after initial release

depthai_sdk/setup.py
```diff
@@ -0,0 +1,36 @@
+import io
+
+from setuptools import setup
+
+with open('requirements.txt') as f:
+    required = f.readlines()
```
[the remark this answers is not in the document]
VanDavv: So on why they are fixed right now
- PyTube was proven to be instable since after one new release we couldn't download YT videos at all
- blobconverter is our tool, so we will try to point to latest version anyways, so fixing it here could update an existing installation if found
- OpenCV had issues with its latest versions not being correctly installed under different OSes, and we wanted to point users to the versions.

`depthai` is not fixed since I didn't want to change library version check right now and it was easier to have it "loose" in SDK and fixed in demo requirements.

I would say to leave it as is for now, since these are the exact requirements we had in the demo script, and for initial release it's ok. We should however check for the cases where opencv might be preinstalled or if we could update/loosen our current requirements.

depthai_sdk/src/depthai_sdk/__init__.py
VanDavv: It's just for convenience - so if someone wants to have everything imported in one line, he still can
```python
from depthai_sdk import to_tensor_result, Previews, PipelineManager
```

Personally, I like to split the declarations into managers/others, like this

```python
from depthai_sdk.managers import PipelineManager
from depthai_sdk import to_tensor_result, Previews
```

However, someone could also be very strict with his imports and do like this

```python
from depthai_sdk.managers.pipeline_manager import PipelineManager
from depthai_sdk.utils import to_tensor_result
from depthai_sdk.previews import Previews
```

And actually, the code snippet above shows how the package would be imported without the imports in `__init__.py`, which allows previous snippets to work.
So it's more for users convenience and allowing them to import the package as preferred

depthai_sdk/docs/Makefile
VanDavv: It was generated automatically so I didn't look into it 😄 will remove

depthai_sdk/docs/source/conf.py
```diff
@@ -0,0 +1,81 @@
+# Configuration file for the Sphinx documentation builder.
```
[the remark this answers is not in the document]
VanDavv: - regarding the logo, it's only on our `Main` docs. The API, Hardware or GUI didn't have them. I think that we should remove the logo from main docs also, since it's causing troubles with rendering, and move it to e.g. navbar - will be much easier to align and will be automatically visible on all docs subprojects
- Yes, it's a TODO, I'll add this to the docs as soon as we release the package

depthai_demo.py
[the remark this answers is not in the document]
themarpe: > and it's something we can do after initial release

Change the coding style you mean?

I'd advise to do it before the first release IMO

depthai_sdk/src/depthai_sdk/__init__.py
themarpe: But isn't `__init__.py` always called when `depthai_sdk` import is called?

Eg:
```
from depthai_sdk.managers import PipelineManager
from depthai_sdk import to_tensor_result, Previews
```

The above would include Previews into `depthai_sdk` right?

Anyway, as long as there aren't name clashes between `managers`, `previews`, `utils`, etc... its probably okay

VanDavv: Yes, but the following import
```
from depthai_sdk.managers import PipelineManager
```
Would not work if `managers/__init__.py` would not contain the proper import, as this class is in `mamagers/pipeline_manager.py` file

depthai_demo.py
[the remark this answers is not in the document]
VanDavv: @themarpe I finished the refactor and everything inside the SDK and demo script is now camelCase

themarpe (changes requested):
Hopefully the changed CLI options names don't affect anything... (and if they do, I don't think those are necessary to be changed)

Also if possible maybe add the sync I've mentioned offline, or atleast to demo itself - it'd be quite nice to have, especially in the demo as it otherwise doesn't show the best capabilities.

Apart from that LGTM

themarpe (approved):

PR #470, opened 2021-08-30
Add `-fps` option to `calibrate.py`

alex-luxonis:
On RPi3 / CM3, by running calibration on all cameras it was observed that image capture (by pressing Space) was unresponsive. Reducing the FPS seems to resolve the issue.


cafemoloko:
@alex-luxonis thanks! So it didn't work with `-fps 4`, so I increased to `-fps 10` and was able to calibrate.
(args used: `-db -brd BW1097 -s 2.4 -fps 10`)

saching13 (approved):
I think we can merge this. 
I will do any additional bug checking on the LITE and will add in that merge

PR #472, opened 2021-09-02
Fixed max (normal) disparity to 95 (0..95)

Erol444:
Also in `config_manager.py` we should remove the helper function `def maxDisparity(self):` and instead use `stereo.getMaxDisparity()`


MOj0 (approved):
LGTM

VanDavv (approved):

PR #473, opened 2021-09-05
fix: for upsidedown image from oak1

auphofBSF:
When using the Oak1 in the `depthai_demo.py` the image is upside down. 

```
 [-cam_model {oakd, oak1}]
```

The following enables a command line argument to allow specifying the model of camera so as to apply appropriate defaults.

See discussion and relevant links https://discuss.luxonis.com/d/311-oak-1-poe-image-image-upside-down-using-depthai-demo-py

and thanks to @alex-luxonis  and Naraka on discord for solution https://discord.com/channels/790680891252932659/799407361986658354/882061206330966117

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,8 @@ python3 install_requirements.py
 
 `python3 depthai_demo.py -vid <path_to_video_or_yt_link>` - CNN inference on video example
 
+`python3 depthai_demo.py -cam_model oak1` - Applies appropriate defaults, ie flips the Camera image Horizontal so that it is correct way up
+
 `python3 depthai_demo.py -cnn person-detection-retail-0013` - Run `person-detection-retail-0013` model from `resources/nn` directory
 
 `python3 depthai_demo.py -cnn tiny-yolo-v3 -sh 8` - Run `tiny-yolo-v3` model from `resources/nn` directory and compile for 8 shaves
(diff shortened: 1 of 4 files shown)


alex-luxonis:
Thank you for the pull request @auphofBSF ! And sorry for the trouble here...

I've created a PR where the depthai library/firmware is modified to auto-detect the OAK-1-PoE device and rotate the camera view accordingly: https://github.com/luxonis/depthai/pull/474. We should have done this a longer time ago...

An option to manually specify the rotation could still be useful, regardless of the device type, if the user finds it better to mount the device upside down (e.g. for cable management in certain applications). 
We could add it, but it would bring a complication for OAK-D devices, when the depth output is used, the calculations would be wrong, as this usecase isn't handled properly yet on firmware side.

auphofBSF:
Thanks @alex-luxonis , I will pull #474  and test.   Would it be useful to change this PR to allow selection by args of any option for setImageOrientation  from dai.CameraImageOrientation.*   ie:  ROTATE_180_DEG.
To handle the Oak-D no support for depth when rotated suggest this ,  
```
# sudo code - depthai_demo  args for --camera_rotation
if oak-d 
   if arg --disable_depth and arg --camera_rotation
        setImageOrientation derived from arg --camera_rotation
   else:
       log.warn "On Oak-d Camera Rotation not supported when running Depth algorithms"
```

VanDavv:
@auphofBSF that's a great suggestion, will definitely be useful if someone wants to use the demo script for cameras in different orientations

VanDavv:
@auphofBSF I've added a PR that allows setting the custom camera orientation - #475

auphofBSF:
#474 works great, thanks @VanDavv. Closing this PR

PR #474, opened 2021-09-06
Fix default camera orientation for OAK-1-PoE, was rotated

alex-luxonis:
The default setting ( `cam_rgb.setImageOrientation(dai.CameraImageOrientation.AUTO)` ) will default now to `ROTATE_180_DEG` (in order to match how the camera module is mounted inside the device).


VanDavv (approved):
Just tested on OAK-1-PoE and works great, thanks @alex-luxonis!

PR #479, opened 2021-09-13
Fix encoding at 3040 resolution mode

VanDavv:
This PR fixes `-rgbr 3040 -bandw low` use case, where RGB camera is ran at 12 MP mode but the output stream gets encoded.

TODO - we currently use only video output stream, where we could use isp or raw outputs too for different preview options

### DIFF
--- a/depthai_helpers/managers.py
+++ b/depthai_helpers/managers.py
@@ -629,38 +629,54 @@ class PipelineManager:
         for xin in filter(lambda node: isinstance(node, dai.node.XLinkIn), vars(self.nodes).values()):
             device.getInputQueue(xin.getStreamName(), maxSize=1, blocking=False)
 
-    def mjpeg_link(self, node, xout, raw_node_output):
+    def __calc_encodeable_size(self, source_size):
+        w, h = source_size
+        if w % 16 > 0:
+            new_w = w - (w % 16)
+            h = int((new_w / w) * h)
+            w = int(new_w)
+        if h % 2 > 0:
+            h -= 1
+        return w, h
+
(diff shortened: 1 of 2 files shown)


depthai_helpers/managers.py
```diff
@@ -629,10 +629,35 @@ def create_default_queues(self, device):
...
     def mjpeg_link(self, node, xout, node_output):
         print("Creating MJPEG link for {} node and {} xlink stream...".format(node.getName(), xout.getStreamName()))
         videnc = self.p.createVideoEncoder()
-        if isinstance(node, dai.node.ColorCamera) or isinstance(node, dai.node.MonoCamera):
+        if isinstance(node, dai.node.ColorCamera):
+            if node.video == node_output:
+                size = self.__calc_encodeable_size(node.getVideoSize())
+                node.setVideoSize(size)
+                videnc.setDefaultProfilePreset(*size, node.getFps(), dai.VideoEncoderProperties.Profile.MJPEG)
+            elif node.preview == node_output:
```
alex-luxonis, line 650: Just a note here: only `video` and `still` from ColorCamera can be encoded, as VideoEncoder needs NV12 format (or grayscale). But maybe at some point we could add an option for ColorCamera preview to output grayscale, then it might be usable (with planar only, and VideoEncoder taking the first plane only if the format is RGB).

alex-luxonis (approved):

depthai_helpers/managers.py
VanDavv: thanks! would it be possible to extend the supported formats? This would allow to encode not only preview stream but also passthrough of depth

PR #481, opened 2021-09-17
remove opencv libraries if exists before installation

VanDavv:


install_requirements.py
```diff
@@ -30,6 +30,7 @@
 if not in_venv:
     pip_install.append("--user")
 
+subprocess.check_call(pip_call + ["uninstall", "-y", "opencv-python", "opencv-contrib-python"]) # remove old versions of OpenCV
```
Erol444, line 33: Should we first check if it's/they're above the required version?

Erol444 (approved):

install_requirements.py
VanDavv: No, it will still succeed and just report a warning that these were not installed

PR #503, opened 2021-10-11
disable overwriting camera matrix

kota-rr:
Disable overwriting camera matrices `M1` and `M2` in mesh generation


kota-rr:
My mistake. This is not necessary. I was confused with ` getOptimalNewCameraMatrix()` function. This parameter is not overwritten and needs to be the same for both L and R.
I will withdraw the PR.

PR #512, opened 2021-10-21
Update requirements installer

szabi-luxonis:
Adds `--prefer-binary` to packages, to prefer precompiled wheels over latest source.
Brings improvements from: https://github.com/luxonis/depthai-python/pull/401

Removed dependencies for opencv-python/contrib specific versions, since SDK already contains them, and we don't hard define opencv version in `depthai-python` neither.

Please test on various platforms.

Why blobconverter version is not set to the latest ? Mainly changes from here: https://github.com/luxonis/depthai/pull/510/files
Pushed hotfix here: https://github.com/luxonis/depthai/commit/a46ad95744d8175f1c87bf8cd92c7423a84b8607

Note, there is an issue between opencv-python and opencv-contrib-python using latest wheels, QT5 related, need to look into it.

### DIFF
--- a/depthai_sdk/requirements.txt
+++ b/depthai_sdk/requirements.txt
@@ -1,4 +1,4 @@
-numpy>1.20
+numpy>1.19
 opencv-python>4
 opencv-contrib-python>4
 blobconverter>=1.2.2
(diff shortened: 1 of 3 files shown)


Erol444:
On RPI: Even after the latest commit (`add numpy to dependency list`), I get this error:
```
from .cv2 import *
ImportError: numpy.core.multiarray failed to import
```
Running install_requirements.py installs numpy `1.16.2`. Running `python3 -mpip install numpy -U` fixes this error (upgrading to 1.21.3).

PR #513, opened 2021-10-22
Introduce QT GUI

VanDavv:
This PR introduces QT-based GUI

Screenshot 2021-11-03 at 11 52 13

You can see a "play-around" video here - https://www.youtube.com/watch?v=sCZpsFQBffk

Some of the options you can see disabled, as changing these requires demo script restart - which will be added (in todos), either in this PR or in a separate one if we'd like to split these

Together with this PR, the demo code itself got refactored - it's now class-based with the main class called `Demo`. This will allow us to import and modify the demo code or specify callbacks easier - but I didn't put too much into this aspect yet as it was not the core aspect of this PR
The change was required since the demo script now runs in a separate thread and shares its memory with main thread.

TODO:
- Tested on Mac, will need to test it on Ubuntu and Windows too before merge
- Investigate FPS limitations (it's pretty poor on video, have to check if it's due to USB speed or QT Signals speed)
- letterboxing instead of frame resize
- enable options that require demo restart


szabi-luxonis (commented):
Tested w/ python 3.6 and 3.10 on ubuntu 21.10, works fine apart from some issues:

Rectified left/right frames are flipped relative to depth frames.

Seems to be placement issues here:
image

VanDavv:
@szabi-luxonis addressed the issues - I didn't see the issue on my side with the placement, but moved these values a bit lower just in case.
Screenshot 2021-11-08 at 10 18 11

szabi-luxonis (approved):

PR #527, opened 2021-11-18
Add demo GIF showing GUI usage

VanDavv:
This PR adds a demo GIF with a link to YouTune showing an example demo run

[depthai demo](https://www.youtube.com/watch?v=sCZpsFQBffk) 

Tried to add a higher-resolution GIF, but GitHub routes all images via https://camo.githubusercontent.com/ (even for external sourced), and if a file exceeds the size limit (10MB) it will raise "Content length exceeded" error and not return an image (404)

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,9 @@
 
 This repo contains demo application, which can load different networks, create pipelines, record video, etc.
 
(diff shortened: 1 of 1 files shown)


Erol444 (approved):
Depth isn't perfect, as the camera is facing walls, so we should re-record with OAK-D-Pro when we get it.

VanDavv:
Actually, I will be recording a new video for this PR - https://github.com/luxonis/depthai-python/pull/390 - so will rerecord this one too then. For now, let's use this one and I'll replace both

PR #540, opened 2021-11-23
Added capability to force update to a new version

themarpe:
Adds capability to force checkout new version even if there were changes in the depthai repository

### DIFF
--- a/launcher/launcher.py
+++ b/launcher/launcher.py
@@ -202,18 +202,28 @@ class Worker(QtCore.QThread):
                             # DepthAI repo not available, clone first
                             splashScreen.updateSplashMessage('Updating DepthAI Repository ...')
                             splashScreen.enableHeartbeat(True)
-                            lastCall = subprocess.run([gitExecutable, 'checkout', '--recurse-submodules ', newVersionTag], cwd=pathToDepthaiRepository, stderr=subprocess.PIPE)
-                            if lastCall.returncode != 0:
+                            lastCall = subprocess.run([gitExecutable, 'status', '--porcelain'], cwd=pathToDepthaiRepository, stdout=subprocess.PIPE)
+                            filesToRemove = lastCall.stdout.decode()
+                            lastCall = subprocess.run([gitExecutable, 'checkout', '--recurse-submodules', newVersionTag], cwd=pathToDepthaiRepository, stderr=subprocess.PIPE)
+                            if lastCall.returncode != 0 or filesToRemove != "":
                                 # Uncommited changes - redo with a prompt to force
+                                # Or unclean working directory
                                 errMessage = lastCall.stderr.decode()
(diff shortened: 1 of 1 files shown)


launcher/launcher.py
```diff
@@ -197,29 +197,45 @@ def run(self):
...
-                                title = 'Update Aborted'
-                                message = f'DepthAI Repository has uncommited changes. Update was aborted.\n{errMessage}'
+                                title = 'Force Update'
+                                message = f'DepthAI Repository has uncommited changes. Do you want to override the changes?\n{errMessage}'
                                 print(f'Message Box ({title}): {message}')
-                                self.sigWarning.emit(title, message)
+                                self.signalUpdateQuestion.emit(title, message)
+
+                                if self.shouldUpdate == True:
+                                    lastCall = subprocess.run([gitExecutable, 'checkout', '-f', newVersionTag], cwd=pathToDepthaiRepository, stderr=subprocess.PIPE)
```
VanDavv, line 215: I think I didn't update anything in my case manually (when receiving the error). So maybe the uncommitted changes are created somehow during installation.

Could you also add `git clean -fx` that would remove the local changes before checkout, this will make sure the directory is clean before checkout

VanDavv (changes requested):

launcher/launcher.py
themarpe: > So maybe the uncommitted changes are created somehow during installation.

I haven't saw that behavior on a installer created from a tagged commit yet (neither did with other installers created from a clean commit)

I wonder what behavior we want to have here. Best case for launcher only usecase is to always do a complete clean&reset to a new commit, so its like a fresh clone of that state. 
The counter point is, one could also use this in repo (local development, etc...) and maybe retaining this git like functionality is better.

Also `git clean -fx` seems to only affect untracked files and not changes to tracked files, as far as I've read into it. Not sure if we want to remove ignored&untracked files?

VanDavv: @themarpe the demo allows both callbacks file and direct import as ways of extending the demo - and these can be done outside of our repository. The same thing goes with model root, it can be specified to point anywhere in the system. 

So I think it's safe to have a clean repo each time, we will be able to still serve the same functionalities and won't deal with dirty repository issues. Also, the installer is designed for less-technical people, so I would assume they want just to experiment with the script, not necessarily augment it

themarpe: I'll perform the clean conditionally as well in that case and only for non-ignored files as I think those don't get changed during regular use case when developing

themarpe: Updated - checks if clean needs to be performed and asks to do it along with force checkout. 
Cleans only files that aren't specified in gitignore.

VanDavv (approved):

PR #542, opened 2021-11-24
Fix Windows signing CI

VanDavv:


.github/workflows/demo.workflow.yml
```diff
@@ -39,8 +36,12 @@ jobs:
   release:
     if: startsWith(github.ref, 'refs/tags/v')
     runs-on: windows-latest
+    needs: windows-installer
     steps:
-
+    - uses: actions/checkout@v2
+      with:
+        submodules: 'recursive'
+        persist-credentials: ''
```
themarpe, line 44: Not needed in release step

themarpe (approved):
Apart from that LGTM
Thanks for testing it out

PR #543, opened 2021-11-24
Fix VideoEncoder deprecated warnings

alex-luxonis:
Some warnings were printed with USB2 / PoE / `-enc <cam>`:
`[warning] VideoEncoder setDefaultProfilePreset: passing 'width'/ 'height' is deprecated. The size is auto-determined from first frame`

Passing the input size was deprecated in depthai 2.13.0.0, the size is auto-configured from the first received frame.


VanDavv (approved):
Thanks! LGTM

Issue #541, opened 2021-11-24
[BUG] Oak-D-PoE on Mac and LAN, is visible, but produces [X_LINK_BOOTLOADER] error


mpurnyn:
**Check if issue already exists**
I checked and nothing looks similar enough to my circumstances. 

**Describe the bug**
I got a new Oak-D-PoE and I wanted to try to connect to it with the demo code. In the past I have managed to get the OAK-D and OAK working on Mac, Windows, and Linux over USB however get this error when trying to do the same over Ethernet. I first tried to connect directly to the PC and I managed to ping it, but wasnt able to find it automatically, but then I tried to do it over LAN and It discovers it, but produces this error.

`(luxonis) ppurnyn@JP-KIT-M-232 depthai % python3 depthai_demo.py 
Using depthai module from:  /Users/ppurnyn/Library/Python/3.9/lib/python/site-packages/depthai.cpython-39-darwin.so
Depthai version installed:  2.13.0.0
Setting up demo...
Available devices:
[0] 192.168.1.112 [X_LINK_BOOTLOADER]
Stack trace (most recent call last) in thread 12920614912:
#11   Object "libsystem_pthread.dylib", at 0x7ff811701513, in _pthread_start + 124
#10   Object "QtCore", at 0x163e556a7, in QTimerInfoList::activateTimers() + 5463
#9    Object "QtQuick", at 0x16a114b27, in QNativeInterface::QSGMetalTexture::fromNative(id<MTLTexture>, QQuickWindow*, QSize const&, QFlags<id<MTLTexture>::CreateTextureOption>) + 28471
#8    Object "QtCore", at 0x163e3c989, in QMacAutoReleasePool::~QMacAutoReleasePool() + 41
#7    Object "Foundation                          0x00007ff812528057", at 0x7ff812528057, in drain] + 132
#6    Object "CoreFoundation", at 0x7ff8117895df, in _CFAutoreleasePoolPop + 21
#5    Object "libobjc.A.dylib", at 0x7ff81159939f, in objc_autoreleasePoolPop + 160
#4    Object "libobjc.A.dylib", at 0x7ff81159c41e, in AutoreleasePoolPage::releaseUntil(objc_object**) + 166
#3    Object "libobjc.A.dylib", at 0x7ff811597e80, in objc_release + 32
#2    Object "libsystem_platform.dylib", at 0x7ff811716e2c, in _sigtramp + 28
[...]

PR #544, opened 2021-11-24
Apply Qt6 GUI workaround also for macOS

alex-luxonis:
Workaround for the crash on https://github.com/luxonis/depthai/issues/541


VanDavv (approved):
Thanks! I think we can set this variable to any OS I think, as we don't use any 3D processing

PR #550, opened 2021-11-25
Allow to specify which render backend to use (-gt / --guiType)

VanDavv:
To enable RPi to work with latest demo, until we migrate to Qt5 we will still use the OpenCV render methods to display camera previews.

It will switch to this mode by default on RPi, and can be enforced with `-gt / --guiType` parameter accepting either `qt`, `cv` or `auto`.

### DIFF
--- a/depthai_demo.py
+++ b/depthai_demo.py
@@ -886,12 +886,15 @@ def runOpenCv():
 
 
 if __name__ == "__main__":
-    is_pi = platform.machine().startswith("arm")
-    use_cv = args.guiType == "cv" or (args.guiType == "auto" and is_pi)
-    try:
-        import PySide6
-    except:
-        use_cv = True
+    use_cv = args.guiType == "cv"
+    if not use_cv:
+        try:
+            import PySide6
+        except:
+            if args.guiType == "qt":
+                raise
+            else:
+                use_cv = True
     if use_cv:
         args.guiType = "cv"
         runOpenCv()
--- a/depthai_helpers/arg_manager.py
+++ b/depthai_helpers/arg_manager.py
@@ -123,7 +123,7 @@ def parseArgs():
                                                                                                                  "If set to \"high\", the output streams will stay uncompressed\n"
                                                                                                                  "If set to \"low\", the output streams will be MJPEG-encoded\n"
                                                                                                                  "If set to \"auto\" (default), the optimal bandwidth will be selected based on your connection type and speed")
-    parser.add_argument('-gt', '--guiType', type=str, default="auto", choices=["auto", "qt", "cv"], help="Specify GUI type of the demo. \"cv\" uses buil-in OpenCV display methods, \"qt\" uses Qt to display interactive GUI. \"auto\" will use OpenCV for Raspberry Pi and Qt for other platforms")
+    parser.add_argument('-gt', '--guiType', type=str, default="auto", choices=["auto", "qt", "cv"], help="Specify GUI type of the demo. \"cv\" uses built-in OpenCV display methods, \"qt\" uses Qt to display interactive GUI. \"auto\" will use OpenCV for Raspberry Pi and Qt for other platforms")
     parser.add_argument('-usbs', '--usbSpeed', type=str, default="usb3", choices=["usb2", "usb3"], help="Force USB communication speed. Default: %(default)s")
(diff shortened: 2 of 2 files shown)


depthai_helpers/arg_manager.py
```diff
@@ -123,6 +123,7 @@ def parseArgs():
                                                                                                                  "If set to \"high\", the output streams will stay uncompressed\n"
                                                                                                                  "If set to \"low\", the output streams will be MJPEG-encoded\n"
                                                                                                                  "If set to \"auto\" (default), the optimal bandwidth will be selected based on your connection type and speed")
+    parser.add_argument('-guit', '--guiType', type=str, default="auto", choices=["auto", "qt", "opencv"], help="Specify GUI type of the demo. \"opencv\" uses buil-in OpenCV display methods, \"qt\" uses Qt to display interactive GUI. \"auto\" will use OpenCV for Raspberry Pi and Qt for other platforms")
```
alex-luxonis, line 126: Typo: `buil-in`

We'd probably have to recommend `"-guit opencv"` to users for some cases (e.g. at the moment simultaneous view of multiple streams). No strong preference, but I'd change the option `opencv` to `cv` just to be easier to type. And maybe also `-guit` to `-gui` or `-gt`, but partial matching for it works already (e.g. `-gui` or `-g`).

depthai_demo.py
```diff
@@ -871,5 +876,18 @@ def guiOnToggleDisparity(self, value):
...
+def runOpenCv():
+    confManager = prepareConfManager(args)
+    demo = Demo()
+    demo.run_all(confManager)
 
-    App().start()
\ No newline at end of file
+
+if __name__ == "__main__":
+    is_pi = platform.machine().startswith("arm")
```
alex-luxonis, line 889: This might match Mac M1 (`"arm64"`), although that could be a good thing (not sure on the PySide6 availability). But anyway, for now in almost all cases we run with Rosetta, so it returns `"x86_64"`.

alex-luxonis (commented):
The "opencv" view opens many streams: `color, nnInput, depth, depthRaw, rectifiedLeft, rectifiedRight`. It works also on RPi, but it's overwhelming for it to display, FPS dropping to about 6. We should revert to have the main two by default: color and depth. Still possible to customize the stream list with `-s ...`

depthai_demo.py
themarpe: @VanDavv 
I'd also try importing the pyside6 and catching on import if not found and automatically downgrading to `opencv` option

Also, might be worth checking doing the same for software backend (not sure if possible, as those looked like runtime errors from an available backend)

depthai_helpers/arg_manager.py
```diff
@@ -123,6 +123,7 @@ def parseArgs():
                                                                                                                  "If set to \"high\", the output streams will stay uncompressed\n"
                                                                                                                  "If set to \"low\", the output streams will be MJPEG-encoded\n"
                                                                                                                  "If set to \"auto\" (default), the optimal bandwidth will be selected based on your connection type and speed")
+    parser.add_argument('-gt', '--guiType', type=str, default="auto", choices=["auto", "qt", "cv"], help="Specify GUI type of the demo. \"cv\" uses buil-in OpenCV display methods, \"qt\" uses Qt to display interactive GUI. \"auto\" will use OpenCV for Raspberry Pi and Qt for other platforms")
```
alex-luxonis, line 126: Typo: `buil-in`

depthai_demo.py
alex-luxonis: @VanDavv So this also doesn't catch `aarch64`, and maybe shouldn't catch `arm64` (e.g Mac M1). Should we properly fix, or remove it, now as PySide6 import is checked?

VanDavv: @alex-luxonis I changed the logic now, not relying on platform but whether PySide6 is importable or not

alex-luxonis (approved):

Erol444:
IMO `auto` (default) should use previous logic, so it will try to use GUI and if it fails, it uses CV. Otherwise, currently, it just crashes on RPI by default, if you don't manually specify `-gt cv`.

VanDavv:
@Erol444 I just tried on my RPi with no Qt installed and it seemed to work as expected - it defaulted to OpenCV render. Will reach out offline to resolve the issue

PR #551, opened 2021-11-25
Fix installing requirements on Windows: `pyrebase` -> `pyrebase4`

alex-luxonis:
Also upgrades `requests` to latest - fixing vulnerabilities: https://github.com/luxonis/depthai/pull/545

Note: after installing `pyrebase4`, if the package `pyrebase` is somehow uninstalled, `pyrebase4` would stop working, needing to be reinstalled with `--force-reinstall`, or removed first - similar issue as with `opencv-python` and `opencv-contrib-python`. This isn't separately handled yet.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,5 @@
-requests==2.11.1
+requests==2.26.0
+pyrebase4==4.5
 argcomplete==1.12.1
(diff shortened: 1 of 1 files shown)


VanDavv:
Just tested, works great, thanks!

VanDavv (approved):

Issue #553, opened 2021-11-25
depthai_demo.py fails with exception and told me to report to developers


diablodale:
Running `depthai_demo.py` with `face-detection-retail-0004` for 12 shaves throws exception and tells me to report to developers.

### Setup

* Microsoft Windows [Version 10.0.19044.1348]
* depthai repo at `main` currently d51f032df0a100f2879ca7318da333fb779cb6ec
* python 3.9.5 64-bit

### Repro

1. Follow instructions from https://docs.luxonis.com/en/latest/pages/tutorials/first_steps/#setup
2. activate the venv
3. Run `python depthai_demo.py`
4. see the default model run correctly
5. change model to `face-detection-retail-0004`
6. click apply restart
7. restart depthai_demo.py if the app completely exits
8. change model to `face-detection-retail-0004` again if the app exited
9. see face model working
10. change shaves to 12
11. click apply restart
12. restart depthai_demo.py if the app completely exits
13. change model to `face-detection-retail-0004` again if the app exited
14. click apply restart
15. download happens in background... then...

### Result

App throws exception. Tells me to report to developers.
Windows dialog also appears.
![image](https://user-images.githubusercontent.com/679350/143485570-18a97f02-963c-4fae-95e7-a55bf0a16fa7.png)

```
(venv) C:\njs\depthai>python depthai_demo.py
Using depthai module from:  C:\njs\depthai\venv\lib\site-packages\depthai.cp39-win_amd64.pyd
Depthai version installed:  2.13.0.0
Setting up demo...
Available devices:
[0] 14442C10C1A1C6D200 [X_LINK_UNBOOTED]
USB Connection speed: UsbSpeed.SUPER
Stopping demo...
=== TOTAL FPS ===
[color]: 28.2
[nnInput]: 23.7
[depthRaw]: 24.9
[depth]: 24.9
[left]: 27.0
[rectifiedLeft]: 23.6
Setting up demo...
[right]: 26.4
Available devices:
[rectifiedRight]: 24.6
[nn]: 23.6
[0] 14442C10C1A1C6D200 [X_LINK_UNBOOTED]
USB Connection speed: UsbSpeed.SUPER
Downloading C:\Users\dale\.cache\blobconverter\face-detection-retail-0004_openvino_2021.4_12shave.blob...
[==================================================]
Done
[...]

PR #554, opened 2021-11-25
Update depthai lib to latest develop

szabi-luxonis:
Solves https://github.com/luxonis/depthai/issues/553


alex-luxonis:
LGTM.

Also removed skipping the version check if the installed library is not a release (behavior was changed in https://github.com/luxonis/depthai/commit/b423c1a), as it could have let the user run with the wrong version.
Skipping version check (if needed) is already possible with `--skipVersionCheck`.

PR #560, opened 2021-12-01
Install numpy version based on python version

szabi-luxonis:
Numpy `1.19.5` is the latest wheel that supports `python3.6`.


Erol444 (approved):
Just started with a fresh install of RPI OS (bullseye) and this seems to work, thanks!

PR #565, opened 2021-12-02
Increase wait time for the demo subprocess to finish

VanDavv:


Erol444 (commented):
Segmentation fault still occurs when you are just playing around with the depthai-GUI every so often. Not sure if this PR addresses any other problem, which is now fixed though.

Erol444 (approved):

Erol444:
From my testing on RPI4, the `RuntimeError: Device not available again after 10 seconds!`  doesn't appear anymore

PR #569, opened 2021-12-03
disable disparity encoding to prevent OOM issue

VanDavv:
When `disparity` was requested with low-bandwidth mode, it caused Out Of Memory error due to too many encoders being created. This PR disables `disparity` stream encoding when in low-bandwidth mode, so that 3 video encoders are created at once

Also, fixes default colormap used in preview manager to use an improved one


Erol444 (approved):
another option would be to limit pool sizes etc. to reduce RAM consumption

PR #573, opened 2021-12-06
Introduce supervisor mode

VanDavv:
This PR introduces supervisor mode to the demo script. Supervisor is responsible to check user's environment and adjust the run options and environment for the demo to work.

As implemented now, the supervisor will try to run a simple logo-only Qt application and terminate it after 10ms. This allows checking if the rendering backend works without disturbing the users' active window.
Next, the script will run itself again as a subprocess, with `--noSupervisor` flag, having options adjusted. In case of Qt run failure, the demo will try to run again but with OpenCV backend. 

All other environment flags and cli options are preserved, so this change will allow using supervised mode the same we use demo currently. Will be testing it now, focusing on RPi and Ubuntu 18.04 as we have known issues there

This PR is built on top of https://github.com/luxonis/depthai/pull/571


depthai_demo.py
```diff
@@ -10,21 +10,23 @@
 from pathlib import Path
 
 import cv2
-os.environ["DEPTHAI_INSTALL_SIGNAL_HANDLER"] = "0"
+
 import depthai as dai
```
szabi-luxonis, line 14: Let's add try except for dependency packages (e.g. Cv2, numpy, depthai).
This can result in failure if the user runs only the demo, without runnimg install requirements beforehand

szabi-luxonis (changes requested):

requirements.txt
```diff
@@ -6,7 +6,7 @@ opencv-python==4.5.1.48 ; platform_machine != "aarch64" and platform_machine !=
 opencv-contrib-python==4.5.4.58 ; platform_machine != "aarch64" and platform_machine != "armv6l" and platform_machine != "armv7l" and python_version == "3.10"
 opencv-contrib-python==4.5.1.48 ; platform_machine != "aarch64" and platform_machine != "armv6l" and platform_machine != "armv7l" and python_version != "3.10"
 opencv-python==4.4.0.46 ; platform_machine == "armv6l" or platform_machine == "armv7l"
-opencv-contrib-python==4.4.0.46 ; platform_machine == "armv6l" or platform_machine == "armv7l"
+opencv-contrib-python==4.1.0.25 ; platform_machine == "armv6l" or platform_machine == "armv7l"
```
szabi-luxonis, line 9: Why? Test code? 
Discrepancy between opencv python and contrib python might cause issues.

depthai_helpers/supervisor.py
```diff
@@ -0,0 +1,62 @@
...
+            label = QLabel()
+            px = QPixmap(str(Path(__file__).parent / "logo.png"))
+            label.setPixmap(px)
+            label.resize(200, 200)
+            label.setWindowFlags(Qt.FramelessWindowHint)
+            label.setAttribute(Qt.WA_TranslucentBackground)
+            label.show()
+            QTimer.singleShot(10, label.close)
+            exit_code = app.exec_()
+            return exit_code == 0
```
themarpe, line 59: Does supervisor have to show anything? Can we check without that? 
Launcher will run the demo as is, will it show the logo again?

VanDavv: I wanted to mimic a regular run to hopefully detect more issues that can happen. But, since it's only 10ms, the logo does not appear at all, users just experience a slight delay in demo startup

depthai_demo.py
VanDavv: just added - https://github.com/luxonis/depthai/pull/573/commits/081d7f3321563da09009378f0605c36c0b6bde2b

VanDavv: Also pointed user to our install instructions in the docs in case something goes wrong - https://github.com/luxonis/depthai/pull/573/commits/f24f10a5505d79d22c25e085e34cb2b5d78209a2

depthai_helpers/supervisor.py
VanDavv: @themarpe I removed this code eventually, as importing Qt modifies the system and can result in a segfault from Qt

Erol444:
Tried it on RPI4, latest Bullseye. After playing around with the GUI and clicking `apply and restart`, app often can't find the device after 10s. It does boot up the CV GUI (which is better than crashing), but we should fix this issue of not finding the device afterward.
```
t.qpa.xcb: QXcbConnection: XCB error: 148 (Unknown), sequence: 192, resource id: 0, major code: 140 (Unknown), minor code: 20
USB Connection speed: UsbSpeed.SUPER
  File "/home/pi/depthai/gui/main.py", line 49, in applyAndRestart
    instance.restartDemo()
  File "/home/pi/depthai/depthai_demo.py", line 682, in restartDemo
    self.stop()
  File "/home/pi/depthai/depthai_demo.py", line 679, in stop
    raise RuntimeError("Device not available again after 10 seconds!")
RuntimeError: Device not available again after 10 seconds!
Aborted
Error while running demo script... Command '/usr/bin/python3 depthai_demo.py --noSupervisor --guiType qt' returned non-zero exit status 134.
Waiting 5s for the device to be discoverable again...

```

Erol444 (approved):
From my testing, it does what it's supposed to do, so that's great. We are still experiencing a sporadic `Segmentation fault` error on RPi4.

PR #577, opened 2021-12-13
Add download progress indicator in Qt GUI

VanDavv:
This PR introduces download progress indicator to Qt GUI, as shown in the demo below

ezgif com-gif-maker (3)

Related blobconverter PR - https://github.com/luxonis/blobconverter/pull/31 - and release - https://github.com/luxonis/blobconverter/releases/tag/v1.2.8

### DIFF
--- a/depthai_demo.py
+++ b/depthai_demo.py
@@ -62,7 +62,7 @@ class Demo:
         self.setup(conf)
         self.run()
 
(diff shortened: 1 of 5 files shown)


VanDavv:
Since we're experiencing some segfaults on Pi, I didn't want to introduce new Qt components until I find out the issue - that's why it's implemented to just send frames same way we send preview frames

PR #578, opened 2021-12-14
exclude latest version of Qt due to Windows issue

VanDavv:
This PR solves the reported issue on Windows, by preventing 5.15.6 from being installed and thus mitigated the issue below

```
  File "D:\depthai\depthai_demo.py", line 893, in <module>
    runQt()
  File "D:\depthai\depthai_demo.py", line 493, in runQt
    from gui.main import DemoQtGui, ImageWriter
  File "D:\depthai\gui\main.py", line 5, in <module>
    from PyQt5.QtQml import QQmlApplicationEngine, qmlRegisterType, qmlRegisterSingletonType, QQmlEngine
ImportError: DLL load failed while importing QtQml: The specified module could not be found.
```


themarpe:
I think it might make sense pinning dependencies, for demo case. SDK can be more relaxed on the other hand

PR #583, opened 2021-12-17
Apps

VanDavv:
This PR introduces an apps mechanism, that allows running the isolated applications (requiring a specific set of requirements or API version) from the demo. The first introduced app is UVC one, enabling users to use their OAK device as a web camera.

To give it a try, run the following command
```
$ python3 depthai_demo.py --app uvc
```

### DIFF
--- a/apps/uvc/requirements.txt
+++ b/apps/uvc/requirements.txt
@@ -1,0 +1,9 @@
+opencv-python==4.5.4.58 ; platform_machine != "aarch64" and platform_machine != "armv6l" and platform_machine != "armv7l" and python_version == "3.10"
+opencv-python==4.5.1.48 ; platform_machine != "aarch64" and platform_machine != "armv6l" and platform_machine != "armv7l" and python_version != "3.10"
+opencv-contrib-python==4.5.4.58 ; platform_machine != "aarch64" and platform_machine != "armv6l" and platform_machine != "armv7l" and python_version == "3.10"
+opencv-contrib-python==4.5.1.48 ; platform_machine != "aarch64" and platform_machine != "armv6l" and platform_machine != "armv7l" and python_version != "3.10"
+opencv-python==4.4.0.46 ; platform_machine == "armv6l" or platform_machine == "armv7l"
+opencv-contrib-python==4.4.0.46 ; platform_machine == "armv6l" or platform_machine == "armv7l"
+-e ../../depthai_sdk
+--extra-index-url https://artifacts.luxonis.com/artifactory/luxonis-python-snapshot-local/
+depthai==2.13.3.0
--- a/apps/uvc/requirements.txt
+++ b/apps/uvc/requirements.txt
@@ -1,3 +1,4 @@
+--extra-index-url https://www.piwheels.org/simple
 opencv-python>4
 -e ../../depthai_sdk
 --extra-index-url https://artifacts.luxonis.com/artifactory/luxonis-python-snapshot-local/
--- a/apps/uvc/requirements.txt
+++ b/apps/uvc/requirements.txt
@@ -1,5 +1,2 @@
---extra-index-url https://www.piwheels.org/simple
-opencv-python>4
--e ../../depthai_sdk
 --extra-index-url https://artifacts.luxonis.com/artifactory/luxonis-python-snapshot-local/
 depthai==2.10.0.0.dev+e8bbeabeed5b6b6ebdfd6f2a71c1b2348b0af44d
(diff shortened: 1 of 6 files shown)


depthai_helpers/app_manager.py
```diff
@@ -0,0 +1,37 @@
...
+        self.appPath = appPath or Path(__file__).parent.parent / "apps" / self.appName
+        self.venvPath = self.appPath / "venv"
+        self.appPip = str(self.venvPath / "bin" / "pip")
+        self.appInterpreter = str(self.venvPath / "bin" / "python")
+        self.appRequirements = appRequirements or self.appPath / "requirements.txt"
+        self.appEntrypoint = appEntrypoint or self.appPath / "main.py"
+
+    def createVenv(self, force=False):
+        try:
+            subprocess.check_call([sys.executable, '-m', 'venv', '-h'], env=initEnv, stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
```
szabi-luxonis, line 21: Need to put all occourences of sys.executable in codebase between quotes, otherwise it won't work if there's a whitespace in path

depthai_helpers/app_manager.py
```diff
@@ -0,0 +1,37 @@
...
+            raise
+
+        if not force and Path(self.appInterpreter).exists():
+            print("Existing venv found.")
+        else:
+            print("Creating venv...")
+            subprocess.check_call(' '.join([sys.executable, '-m', 'venv', str(self.venvPath)]), env=initEnv, cwd=self.appPath)
+        print("Installing requirements...")
+        subprocess.check_call(' '.join([self.appPip, 'install', '-U', "pip"]), env=initEnv, shell=True, cwd=self.appPath)
+        subprocess.check_call(' '.join([self.appPip, 'install', '-r', str(self.appRequirements)]), env=initEnv, shell=True, cwd=self.appPath)
```
szabi-luxonis, line 33: Hmm, maybe add --prefer-binary flag? Same as for "main" app requirements installer.

apps/uvc/requirements.txt
```diff
@@ -0,0 +1,5 @@
+--extra-index-url https://www.piwheels.org/simple
+opencv-python>4
```
szabi-luxonis, line 2: Uvc app doesn't require opencv, only that specific depthai version

szabi-luxonis (commented):
Looks like it doesn't work if I specify python version explicitly, e.g. 
`python3.6 depthai_demo.py --app uvc`

VanDavv:
@szabi-luxonis the issue is missing `python3.6-venv` module, just added a prompt that asks you to install this dependency if creating a new venv fails.

apps/uvc/requirements.txt
```diff
@@ -0,0 +1,2 @@
+--extra-index-url https://artifacts.luxonis.com/artifactory/luxonis-python-snapshot-local/
+depthai==2.10.0.0.dev+e8bbeabeed5b6b6ebdfd6f2a71c1b2348b0af44d
```
szabi-luxonis, line 2: Need to update to the one mentioned by Alex `depthai==2.13.3.0.dev+b9dbc05ea35d19534678d791f8fbf03ec4b087ba`
Otherwise LGTM, let's merge and add a `temporary` check for windows, that it's not working and document/check for libraries that cause issues on linux system: `sudo apt remove uvcdynctrl uvcdynctrl-data`. It will create huge log files otherwise.

szabi-luxonis (approved):

apps/uvc/requirements.txt
VanDavv: @szabi-luxonis I would add the check for libraries along with Ubuntu 21.04 fix (with `python3-pyqt5.sip` module) to the `install_dependencies.sh` script, as it should be responsible for preparing the whole system. Updated the version and added a temporary disabled message in the script itself, thanks!

szabi-luxonis: Yes, but older users will not run the script again. 
Either way I agree that we need to add to install dependencies

VanDavv: @szabi-luxonis PR with update is here - https://github.com/luxonis/depthai-python/pull/463
> older users will not run the script again
I think we should recommend installing it each time some error happens, just to make sure all of the updated packages are there - as it can be run multiple times in a row without issues

szabi-luxonis: Yeah, we should 👍🏻 
In this case it's not a hard error, it will just write tens of GBs of log file to the disk.

VanDavv: Actually, it was hard in my case, as I ran out of space available on my Jetson when this package was installed 😀  and couldn't run anything until I deleted this log file

depthai_helpers/app_manager.py
themarpe: @VanDavv @szabi-luxonis 
Check call AFAIK takes arguments as is - the quoting here shouldn't make a difference. If `sys.executable` has spaces it'll still be sent correctly as "argv0".

Quoting is a shell concept to correctly determine what goes to eg "argv0", "argv1", ...

szabi-luxonis: Indeed, the subprocess is not called in shell context, e.g. `shell=True` so it's not needed 👍🏻

apps/uvc/requirements.txt
VanDavv: We can actually keep things the same (to use `install_requitements.py` still) and execute `install_dependencies.(sh|bat)` inside (as a subprocess call). Since it's safe to run this multiple times, there shouldn't be any downside to it. Will address in a different PR

PR #585, opened 2021-12-20
Add sentry integration

VanDavv:
This PR adds sentry integration allowing us to capture runtime errors from our users

### DIFF
--- a/depthai_demo.py
+++ b/depthai_demo.py
@@ -46,8 +46,7 @@ try:
     import sentry_sdk
 
     sentry_sdk.init(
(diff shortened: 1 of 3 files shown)


themarpe:
Will this replace firebase?

szabi-luxonis (approved):
LGTM

VanDavv:
> Will this replace firebase?
@themarpe it depends on how we'll want to use Sentry. For now, I'd stick with both, as I didn't find an easy way to setup some "custom events" in sentry. One thing I'll investigate further is "Crash Free Sessions" feature, that may enable us to collect the same metrics we do now in Firebase

PR #587, opened 2021-12-21
add fix for Windows paths

VanDavv:
This PR fixes the WIndows issue when the path to the interpreter was containing spaces - e.g. `C:\Program Files`. Now, the path is quoted to prevent it

### DIFF
--- a/depthai_helpers/supervisor.py
+++ b/depthai_helpers/supervisor.py
@@ -33,7 +33,7 @@ class Supervisor:
             new_env["LD_LIBRARY_PATH"] = str(Path(importlib.util.find_spec("PyQt5").origin).parent / "Qt5/lib")
             new_env["DEPTHAI_INSTALL_SIGNAL_HANDLER"] = "0"
             try:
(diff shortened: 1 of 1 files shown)


szabi-luxonis (approved):

depthai_helpers/supervisor.py
```diff
@@ -33,7 +33,7 @@ def runDemo(self, args):
             new_env["LD_LIBRARY_PATH"] = str(Path(importlib.util.find_spec("PyQt5").origin).parent / "Qt5/lib")
             new_env["DEPTHAI_INSTALL_SIGNAL_HANDLER"] = "0"
             try:
-                subprocess.check_call(' '.join([sys.executable, "depthai_demo.py"] + new_args), env=new_env, shell=True, cwd=repo_root)
+                subprocess.check_call(' '.join([f'"{sys.executable}"', "depthai_demo.py"] + new_args), env=new_env, shell=True, cwd=repo_root)
```
themarpe, line 36: Any specific reason for using the shell and having a string of args instead of list of args?

PR #589, opened 2021-12-23
Add ability to run apps from GUI

VanDavv:
This PR introduces the ability to run Demo Apps from Qt GUI and then terminate them when we're done playing with the app

Usage demo

https://user-images.githubusercontent.com/5244214/147253793-23820ed0-4fdb-4ecb-8217-818239535e15.mov

### DIFF
--- a/depthai_demo.py
+++ b/depthai_demo.py
@@ -33,7 +33,7 @@ from depthai_helpers.arg_manager import parseArgs
 from depthai_helpers.config_manager import ConfigManager, DEPTHAI_ZOO, DEPTHAI_VIDEOS
 from depthai_helpers.metrics import MetricManager
 from depthai_helpers.version_check import checkRequirementsVersion
-from depthai_sdk import FPSHandler, loadModule, getDeviceInfo, downloadYTVideo, Previews, resizeLetterbox
+from depthai_sdk import FPSHandler, loadModule, getDeviceInfo, downloadYTVideo, Previews, createBlankFrame
 from depthai_sdk.managers import NNetManager, PreviewManager, PipelineManager, EncodingManager, BlobManager
 
 args = parseArgs()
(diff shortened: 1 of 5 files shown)


szabi-luxonis (approved):
Nice!

themarpe:
:+1: 
@VanDavv can you create a separate tab for these? 
(or maybe later on when more apps arrive)
Could also maybe be named "Pipelines", for more fully featured examples to "coin the term" - this would signify a specific workload/problem that the "app" solves, eg. People Counting Pipeline, ... (note: not to be confused with device side pipeline of `depthai-core`)

themarpe (commented):
Any thoughts on "apps" that would like to use GUI? Or will those be integrated and will have to use same core version?

VanDavv:
My understanding of the apps is that they enable features that are easier to run from a script file, rather than a demo - so UVC, mjpeg, webrtc etc would work now.

For apps that have a preview - this is more tricky, as we're running these apps in a separate process, so communication between these is limited. I hope that we will be able to use CV preview in these, didn't check it yet but it might cause issues as QT is loaded at the same time. Will investigate it further a bit later (would like to add other features first and put focus to testing on different platforms)

Regarding "people counting pipeline" and other custom pipelines - this might be actually other feature of the demo we can use in parallel with apps. Since the demo script is modular, we could potentially replace both `setup` and `run` functions with different ones and still use the demo in the same manner. So, implementing pipelines like people counting could be done directly in the demo itself. But this is just an idea and it might turn out to be more difficult to implement than it seems, we can discuss this approach further

VanDavv:
And since we have just a single app, I didn't create a separate tab for them yet. With current placement it should be enough for 4-5 apps to fit, and if we get to this point we can create a separate space just for them

VanDavv:
@themarpe let's discuss this further offline, will merge this PR for now and if we decide to change/rework the apps, will address in a separate PR

PR #590, opened 2021-12-24
remove emotions recognition network (temporarily)

VanDavv:
This PR removes emotions recognition network from resources/nn folder. Since we don't have a 2 stage inference enabled in the demo yet, this cause only confusion as the network is not runnable.

I already created a new branch - https://github.com/luxonis/depthai/tree/add_two_stage_inference - where I'll add this feature and where the emotions recognition network is still available


szabi-luxonis (approved):
Lgtm
Anything else we can remove now that we use blobconverter?

VanDavv:
@szabi-luxonis I think we will have to keep the rest of the nn files despite blobconverter usage, at least until we can get the network input and ov version straight from the blob. Even then we still need to know the type of the network (mobilenet/yolo/etc) and its class labels - but the network would be runnable. For now, users can provide these required attributes using cli args to run their custom blob

themarpe:
@VanDavv 

> Even then we still need to know the type of the network (mobilenet/yolo/etc) and its class labels - but the network would be runnable.

Should we provide this "metadata" along the actual blobs? With upcoming model-zoo of our own, I think it would make even more sense. And this metadata would just be a default - if a user wants to override the labels, it'd be easy to do so, etc...

VanDavv:
Yes, I was thinking about the BlobConverter API to seek for the NN metadata and return them as response headers with the compiled blob - so we can consume it nicely in CLI and return along the blob itself as config dict. 
Our model zoo will need the OMZ models metadata too (which I guess we'd have to prepare ourselves, like labels or network type), so these two efforts fit nicely together

PR #591, opened 2021-12-24
Sync previews

VanDavv:
This PR introduces `syncPreviews` mode that allows to show all frames in sync - meaning they have the same sequence number, hence it can be assumed they were captured at the same time.

To enable from GUI, open "Camera" tab and click "Sync Previews" switch

Screenshot 2021-12-24 at 12 37 46

In CLI, you can use `--syncPreviews` flag to enable this feature.

This PR also removes `--sync` flag, as it only made the `color` stream to show `nnInput` - and we have both of these enabled by default in Qt GUI, and `nnInput` can be requested additionally in CV mode too. Will work on adding `--syncNN` flag that will make sync between nn results and preview frames


themarpe (changes requested):
As discussed offline - lets make the "sync" default and add only "low latency" switch which skips the syncing and ordering of frames and keeps pushing them out to be displayed ASAP.

This is from user perspective better UX as that will be the setting they are looking for if they don't like the slight latency increase which should be the only byproduct of this improvement

PR #592, opened 2021-12-24
Add keyboard interrupt catch blocks

VanDavv:
This PR adds handles for KeyboardInterrup errors, exiting the program gracefully now

### DIFF
--- a/depthai_demo.py
+++ b/depthai_demo.py
@@ -13,10 +13,13 @@ import platform
 from depthai_helpers.app_manager import App
 if __name__ == "__main__":
     if '--app' in sys.argv:
-        app = App(appName=sys.argv[sys.argv.index('--app') + 1])
-        app.createVenv()
-        app.runApp()
-        sys.exit(0)
+        try:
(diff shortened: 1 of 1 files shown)


themarpe (approved):
LGTM if these are the highest possible levels of handling this exception (eg. wrap the entry which will take care if this exception is raised at any level deeper)

themarpe:
Also make sure that `sys.exit(0)` is the right thing to do. What if we are saving a file? Does GUI exit cleanly? ...

VanDavv:
> is the right thing to do. What if we are saving a file? Does GUI exit cleanly? ...
@themarpe this is all handled by context managers (or at least should be). And actually, system exit is another exception - `SystemExit` - and can even be caught, so I think it will go up through the call stack and exit gracefully, same as the KeyboardInterrupt does, but without traceback and with different exit code.
I'll gave it a try a couple of times and worked as expected, but didn't test it yet with encoding enabled, will test if that behaves as it should

themarpe:
Thanks - if it acts that way then that is great - thought it was more of an "exit" equivalent, which just exits the process ungracefully.

PR #600, opened 2022-01-05
Optimize RPi performance

VanDavv:
This PR improves performance on RPi, by removing the preview encoding and moving preview scaling to device

Screenshot 2022-01-05 at 13 13 54

This PR also contains changes from #591 , which changed a bit the way frames are processed (and added `--syncPreviews` flag). We can either merge this PR and have both of these features ready, or merge each one separately (the optimizations without sync mode are here - https://github.com/luxonis/depthai/pull/601) - CC @themarpe @szabi-luxonis


szabi-luxonis (approved):
Wasn't able to test it, but LGTM if works as expected.
Let's merge it and make a release.

PR #611, opened 2022-01-14
Update depthai to 2.15.3.0; Enable RGB-Depth alignment by default

VanDavv:
This PR adds RGB-Depth alignment to the demo and enables LR check by default. To enable this feature, use the `--rgbDepthAlign` flag or click the `RGB-Depth Alignment` switch in Camera properties

Screenshot 2022-01-19 at 14 14 26


szabi-luxonis:
@VanDavv updated with latest imrpovements. If all good from your POV can be merged. 
The GUI doesn't work for me, even on `main`. Tried on 2 different systems, after install requirements.

VanDavv:
Thanks @szabi-luxonis! Will take a look at it and fix, didn't check this branch lately so probably need to add some updates

PR #612, opened 2022-01-14
Add improvements to sync mode

VanDavv:
This PR adds improvements for sync mode, removing DelayedQueue as it adds an unnecessary freeze (delay itself is caused by iteration loop so we don't need an extra one)


themarpe (approved):
The simpler the better - LGTM if functionality wise is the same
(I have to test it out once, still haven't gotten around to it yet...)

szabi-luxonis (approved):

Issue #616, opened 2022-01-19
[BUG] UVC demo app does not work with OAK-D-Lite


zegkljan:
**Describe the bug**
Running `python3 demo_script.py --app uvc` with OAK-D-Lite creates /dev/video* device(s) only for a few seconds, the camera then disconnects and reconnects again as a non-uvc device, the /dev/video* devices disappear.

**To Reproduce**
(interlaced with console and dmesg outputs)
1. Connect OAK-D-Lite

`dmesg`:

```
[ 3283.174698] usb 1-1: new high-speed USB device number 36 using xhci_hcd
[ 3283.315643] usb 1-1: New USB device found, idVendor=03e7, idProduct=2485, bcdDevice= 0.01
[ 3283.315665] usb 1-1: New USB device strings: Mfr=1, Product=2, SerialNumber=3
[ 3283.315674] usb 1-1: Product: Movidius MyriadX
[ 3283.315681] usb 1-1: Manufacturer: Movidius Ltd.
[ 3283.315687] usb 1-1: SerialNumber: 03e72485
```

`v4l2-ctl --list-device`:
```
v4l2loopback (platform:v4l2loopback-000):
	/dev/video9

HD WebCam: HD WebCam (usb-0000:00:14.0-7):
	/dev/video0
	/dev/video1
	/dev/media0
```

2. Run `python3 depthai_demo.py --app uvc`

`python3 depthai_demo.py --app uvc` console:
```
Existing venv found.
Installing requirements...
Requirement already satisfied: pip in ./venv/lib/python3.10/site-packages (21.3.1)
Looking in indexes: https://pypi.org/simple, https://artifacts.luxonis.com/artifactory/luxonis-python-snapshot-local/
Requirement already satisfied: depthai==2.13.3.0.dev+f513b0f27b4f07a9940cd9dfe65d7fd6deeaf29d in ./venv/lib/python3.10/site-packages (from -r /home/zegkljan/tmp/depthai/apps/uvc/requirements.txt (line 2)) (2.13.3.0.dev0+f513b0f27b4f07a9940cd9dfe65d7fd6deeaf29d)
[18443010312C7C0F00] [89.076] [system] [info] PRINT:LeonCss: bss_init took 42.307 ms
BootloaderConfig.options1 checksum doesn't match. Is: 0x10304418 should be: 0x7018171A
BootloaderConfig.options2 checksum doesn't match. Is: 0x00214442 should be: 0x11FD80E1
[...]

PR #618, opened 2022-01-24
Enable CLI params in apps, tweak app setup and termination and disable UVC app with USB2 connections

VanDavv:
in this PR:
- Setup and terminate process is modified to correctly handle different OSes (tested on Windows and Mac)
- Removed argcomplete dependency (as we're moving more into GUI and having it listed would require demo apps to have it as well)
- Enabled same CLI arguments parsing for the apps as it is now for the demo
- Disabled UVC app is connection speed is USB2 due to #616

### DIFF
--- a/apps/uvc/main.py
+++ b/apps/uvc/main.py
@@ -31,8 +31,12 @@ else:
 uvc = pipeline.createUVC()
 cam_rgb.video.link(uvc.input)
 
+
 # Pipeline defined, now the device is connected to
 with dai.Device(pipeline, usb2Mode=args.usbSpeed == "usb2") as device:
+    if device.getDeviceInfo().desc.protocol == dai.XLinkProtocol.X_LINK_USB_VSC and device.getUsbSpeed() not in (dai.UsbSpeed.SUPER, dai.UsbSpeed.SUPER_PLUS):
+        print("This app is temporarily disabled with USB2 connection speed due to known issue. We're working on resolving it. In the meantime, please try again with a USB3 cable/port for the device connection")
+        raise SystemExit(0)
     print("\nDevice started, please keep this process running")
     print("and open an UVC viewer. Example on Linux:")
     print("    guvcview -d /dev/video0")
--- a/apps/uvc/main.py
+++ b/apps/uvc/main.py
@@ -2,13 +2,14 @@
 import platform
 import depthai as dai
 import time
+import sys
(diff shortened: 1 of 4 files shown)


apps/uvc/main.py
```diff
@@ -28,8 +31,12 @@
 uvc = pipeline.createUVC()
 cam_rgb.video.link(uvc.input)
 
+
 # Pipeline defined, now the device is connected to
-with dai.Device(pipeline) as device:
+with dai.Device(pipeline, usb2Mode=args.usbSpeed == "usb2") as device:
+    if device.getDeviceInfo().desc.protocol == dai.XLinkProtocol.X_LINK_USB_VSC and device.getUsbSpeed() not in (dai.UsbSpeed.SUPER, dai.UsbSpeed.SUPER_PLUS):
+        print("This app is temporarily disabled with USB2 connection speed due to known issue. We're working on resolving it. In the meantime, please try again with a USB3 cable/port for the device connection")
+        raise SystemExit(0)
```
szabi-luxonis, line 39: Other than 0? 0 means success usually.

apps/uvc/main.py
```diff
@@ -28,8 +31,12 @@
 uvc = pipeline.createUVC()
 cam_rgb.video.link(uvc.input)
 
+
 # Pipeline defined, now the device is connected to
-with dai.Device(pipeline) as device:
+with dai.Device(pipeline, usb2Mode=args.usbSpeed == "usb2") as device:
+    if device.getDeviceInfo().desc.protocol == dai.XLinkProtocol.X_LINK_USB_VSC and device.getUsbSpeed() not in (dai.UsbSpeed.SUPER, dai.UsbSpeed.SUPER_PLUS):
+        print("This app is temporarily disabled with USB2 connection speed due to known issue. We're working on resolving it. In the meantime, please try again with a USB3 cable/port for the device connection")
```
szabi-luxonis, line 38: Just bringing up:
We could use colored prints for different levels, like in gen1. If there's a library that support windows then that'd great.

szabi-luxonis (approved):

apps/uvc/main.py
VanDavv: @szabi-luxonis I changed it so now we send these print statements to `stderr` output - this way we have it cross-platform and visible better, thanks!

PR #623, opened 2022-01-28
Installer improvements

themarpe:
- Added versioning to installer that matches depthai demo
- Added logging of powershell scripts that install some requirements during installation process 
- Enabled logging by default and added a troubleshooting section to describe where to find the logs

Note: By default, the version is set to `1.0-dev`. For non-release action builds it gets set to `1.0-[commit hash]` and `[version]` for releases


VanDavv (approved):
LGTM

PR #639, opened 2022-02-15
Remove "auto" override of bandwidth argument

themarpe:
Setting `--bandwidth low` or `--bandwidth high` manually, didn't work before

### DIFF
--- a/depthai_demo.py
+++ b/depthai_demo.py
@@ -602,7 +602,6 @@ def runQt():
             self.running = True
             self.signals.setDataSignal.emit(["restartRequired", False])
             self.instance.setCallbacks(shouldRun=self.shouldRun, onShowFrame=self.onShowFrame, onSetup=self.onSetup, onAppSetup=self.onAppSetup, onAppStart=self.onAppStart, showDownloadProgress=self.showDownloadProgress)
-            self.conf.args.bandwidth = "auto"
             if self.conf.args.deviceId is None:
                 devices = dai.Device.getAllAvailableDevices()
                 if len(devices) > 0:


VanDavv:
@themarpe setting bandwidth to auto allows the demo script to determine the actual connection type, which is useful when switching e.g. between different OAKs on different USB ports or between USB and PoE devices.

This param works when used with `--guiType cv`, for QT we can add a switch in GUI that will either set to auto or "low" / "high"

themarpe:
@VanDavv 
It seems like this line overrode the actual choice. The default from arg manager is already `auto`

VanDavv:
@themarpe each time the demo is started, function [`adjustParamsToDevice`](https://github.com/luxonis/depthai/blob/main/depthai_helpers/config_manager.py#L166) is called that makes sure the params are compatible with the specified device - and among these values, `bandwidth` option is overridden.

And since it only triggers when `auto` is set, if you run the QT GUI once with USB3 connected DepthAI, it will set the bandwidth to `high`. If you change the device to a PoE device, the bandwidth parameter will be set to high on the next run, so it won't trigger the low-bandwidth pipeline (e.g., with mjpeg links).

I can change the behavior of the QT GUI app to set the `auto` value for bandwidth only when the device is changed, so this would both mitigate the issue of different connection types and allow to run with specified bandwidth while using a single device

Thoughts @themarpe ?

themarpe:
@VanDavv 
This UX seems odd to me.
I'd have it like such:
 - On each connection to a device, use "auto" IFF its not overriden (set manually by cmd param). If its USB do high bandwidth if its PoE do low, etc...
 - if the parameter is forced manually through command line (eg `--bandwidth low`), use this above anything else for all devices

VanDavv:
@themarpe I will handle it, probably adding commits to this PR. One idea - what do you think about changing `--bandwidth low` to simply `--lowBandwidth`? I assume the main reason to use this flag is to set low bandwidth on high-bandwidth connections, not the other way around.

themarpe:
@VanDavv thanks
I like the way as it is now, seems intuitive to me (also to force the option, in `--lowBandwidth` you'd need `--highBandwidth` as well to force it for eg. PoE devices)

VanDavv:
@themarpe I added a tweak for this in https://github.com/luxonis/depthai/pull/645 . I'll add a better handling for preview there too to solve https://github.com/luxonis/depthai/pull/640 (so that text normalizing won't be necessary to display the data)

themarpe:
> I'll add better handling for preview there too, to solve https://github.com/luxonis/depthai/pull/640 (so that text normalizing won't be necessary to display the data)

Sounds great - I'll close my PRs then

(one idea on above is to resize the frame to be previewed to some constant size and then always draw the same way. Negative is that its more taxing on host to do so)

PR #640, opened 2022-02-15
WIP: normalized text size rendering

themarpe:
@VanDavv 
`fps.py` is in an okay state, while other parts have to still be improved upon.
Mostly difference between frame sizes incoming for PoE vs USB devices (`color` stream) - might be better easier to do it "statically" for the bigger incoming frame size


themarpe:
@VanDavv 
Also, on Linux, the image doesn't stretch the whole window area once it reaches some size.
Do you happen to know why?

PR #645, opened 2022-02-22
Add minor tweaks to demo

VanDavv:


VanDavv:
After the update, this PR allows to use video stream with scaling but ImageManip (probably) is producing artifacts on the final image, hence leaving it still as a draft. The new version was available thanks to @OanaMariaVatavu

PR #656, opened 2022-03-10
Fixed rgb measured translation

saching13:


alex-luxonis (approved):
Just a note that we'll need to add a firmware update to workaround incorrect RGB translation relative to left/right, for already calibrated boards that used the manual calibration - `calibrate.py`, where RGB is not in the center, and RPi CM3/CM4 based: OAK-D-CM3, OAK-D-CM4 and OAK-D-CM4-PoE (1097 and 2097).

Mainly RGB-depth alignment on these boards is affected currently (before adding the FW workaround).

OAK-D-IoT-40 is not affected, as the issues canceled out: incorrect `left-to-rgb-distance` and the computation of `rgb-right` baseline when writing EEPROM data.

PR #659, opened 2022-03-17
Sdk tutorial

ZigiMigi:


depthai_sdk/docs/source/blob_nnet_manager.rst
```diff
@@ -0,0 +1,76 @@
...
+``Blob manager`` is very easy and straight forward to use. We declare it and pass which project we want to use as it's argument.
+
+.. literalinclude:: ./examples/code_fractions/blob.py
+   :language: python
+   :linenos:
+
+After that, the ``blob`` is stored in our variable and we can then pass it to our ``Nnet manager``, as we will se bellow.
+Of course the ``face-detection-retail-0004`` is not the only project that we can use. All other projects that are supported are:
+
+.. include::  blob_projects.rst
```
VanDavv, line 22: @ZigiMigi BlobManager supports all models in both [Open Model Zoo](https://github.com/openvinotoolkit/open_model_zoo) and [our model zoo](https://github.com/luxonis/depthai-model-zoo). 
By using `configPath` you can download any arbitrary model and `zooDir` param you can specify a path to a custom model zoo as well (which can contain both already compiled blobs and model yml files)

BlobManager is able to both reuse existing blobs, download models from model zoo and compile custom models based on yml config

depthai_sdk/docs/source/blob_nnet_manager.rst
```diff
@@ -0,0 +1,76 @@
...
+
+Nnet manager
+------------
+
+``Blob manager`` is a class that is made to help you with setting up neural networks.
```
VanDavv, line 27: typo, nnet manager. This description should be extended, as NNetManager is responsible for all NN-related efforts, including creating a correct pipeline as well as drawing nn results on frames

depthai_sdk/docs/source/blob_nnet_manager.rst
```diff
@@ -0,0 +1,76 @@
...
+
+Getting started
+^^^^^^^^^^^^^^^
+
+As the ``Blob manager`` and ``Nnet manager`` work together, we will need both to use our neural network. We will also need to use the ``Pipeline manager`` and the ``Preview manager`` to make everything work together.
```
VanDavv, line 32: This is false, as BlobManager and NNetManager can (and often are) used separately. 
BlobManager is useful if you want to have the ability to download models based on model config or custom model zoo.
Nnet Manager only requires blob path, which can be provided both from BlobManager but also directly from `blobconverter` module.

I often do `blobPath=blobconverter.from_zoo(...)` as it's simpler

VanDavv: Regarding PipelineManager and PreviewManager connections:
- PipelineManager is able to integrate with `NNetManager`, as it's needed to both add neural network node to the pipeline (`addNN` method) and we need to sync OpenVINO version between nn and pipeline (hence `setNnManager` method). This is one way sync and is mostly optional
- NNetManager is able to integrate with PreviewManager for convenience, but it's also able to parse raw frames passed to it.

With the comments above, my point is that we should show our users that these manager/helper classes are flexible and modular, allowing to pick only the ones you find most suitable

depthai_sdk/docs/source/blob_nnet_manager.rst
```diff
@@ -0,0 +1,76 @@
...
+Face detection
+^^^^^^^^^^^^^^
+
+.. literalinclude:: ./examples/face_detection_color.py
+   :language: python
+   :linenos:
+
+In this above example we will use all above mentioned classes and run the face detection project.
+First we define the pipeline and initialize the streams. After that we load in our blob (``face-detection-retail-0004``) and send it in to our ``Nnet manager``.
+Every project has its own ``inputSize`` and ``familyName``. To see more about the different requirements check the ``depthai/resources/nn`` folder. Every blob has its requirements in it's json file.
```
VanDavv, line 51: `depthai/respources/nn` are specific to depthai demo, it's also not needed when dealing with mobilenet-ssd-based networks (like the one you used), all parameters can be supplied on init.
Maybe nitpicking, but these json files store decoding config, not actual requirements

depthai_sdk/docs/source/encoding_manager.rst
```diff
@@ -0,0 +1,33 @@
...
+Getting started
+---------------
+
+Same as ``Preview manager``, ``Encoding manager`` works hand in hand with the ``Pipeline manager``, so before you can use the encoder, you will first have to declare and initialize the ``Pipeline manager``.
+
+.. literalinclude:: ./examples/code_fractions/encoder.py
+   :language: python
+   :linenos:
+
+The most important part, when using the ``Encoding manager`` is that the encoder must be initialized before the pipeline.
```
VanDavv, line 16: Encoder must be created during pipeline creation, as it adds nodes to the pipeline that gets later executed. So it's best to create encoding manager after PipelineManager, but before sending it to the device

depthai_sdk/docs/source/examples/camera_preview.py
```diff
@@ -0,0 +1,33 @@
...
+# creating color source
+pm.createColorCam(xout=True)
+pm.createLeftCam(xout=True)
+pm.createRightCam(xout=True)
+pm.createDepth(useDepth=True)
+
+# connecting to the device
+with dai.Device(pm.pipeline) as device:
+    # define configs for above sources
+    pv = PreviewManager(display=[Previews.color.name, Previews.left.name, Previews.right.name, Previews.depthRaw.name])
```
VanDavv, line 21: I think it's better to use `depth` instead of `depthRaw`, as it looks better

depthai_sdk/docs/source/examples/code_fractions/encoder.py
```diff
@@ -0,0 +1,27 @@
...
+from depthai_sdk.managers import PipelineManager
+import depthai as dai
+
+
+# Before declaring the pipeline and encoder, we must create a dictionary of streams, with fps ans values
+encodeConfig = dict()
+encodeConfig["color"] = 30
+
+# create encoder with above declared dictionary and path to save the file ("" will save it next to the program file)
+em = EncodingManager(encodeConfig, Path(""))
```
VanDavv, line 12: please use `Path(__file__)`

depthai_sdk/docs/source/examples/code_fractions/previews.py
```diff
@@ -0,0 +1,21 @@
...
+# connecting to the device
+with dai.Device(pm.pipeline) as device:
+    # define configs for above sources
+    pv = PreviewManager(display=[Previews.color.name])
+
+    # create stream queues
+    pv.createQueues(device)
+
+    # prepare and show streams
+    pv.prepareFrames()
```
VanDavv, line 20: this needs to be in a loop

VanDavv (changes requested):
Thanks! They look great! 
I left some comments but overall this is a very good starting point we can iterate further

depthai_sdk/docs/source/nnet_manager.rst
```diff
@@ -0,0 +1,52 @@
...
+If our face is shown, our neural network detects it, but if we cover it, our neural network will not detect it.
+
+Mobile net
+----------
+
+.. literalinclude:: ./examples/mobile_net_ssd.py
+   :language: python
+   :linenos:
+
+This example shows how to use the ``mobile net`` project. The code should be almost the same as the one that we used in the above example, with the only difference being
```
VanDavv, line 47: Full name of this network is `MobileNetSSD`

depthai_sdk/docs/source/pipeline_manager.rst
```diff
@@ -0,0 +1,39 @@
+================
+Pipeline manager
+================
+
+Pipeline is the first class that we will learn how to use as it is the one that is goes hand in hand with every other class.
```
VanDavv, line 5: *PipelineManager

depthai_sdk/docs/source/preview_manager.rst
```diff
@@ -0,0 +1,36 @@
...
+Preview manager
+===============
+
+``Preview manager`` is a class that is made to help you with displaying previews / streams from OAK cameras.
+
+Getting started
+---------------
+
+``Preview manager`` works hand in hand with the ``Pipeline manager``, so before you can use Preview, you will first have to declare and initialize the ``Pipeline manager``.
+But of course you will also have to import bot names to use them.
```
VanDavv, line 11: *both

VanDavv (changes requested):
Overall it's a very good direction, thanks @ZigiMigi !

depthai_sdk/docs/source/indexNEW.rst
```diff
@@ -0,0 +1,61 @@
+================================
```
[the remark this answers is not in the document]
ZigiMigi: We do not actualy, I just left the old index (the one that was beeing used before i started working on this) and made the new one as indexNew.

depthai_sdk/docs/source/index.rst
```diff
@@ -1,34 +1,61 @@
-DepthAI SDK documentation
-=========================
+================================
+Getting started with DepthAI SDK
+================================
+In this tutorial, we’ll show you how to use DepthAI SDK for a couple of basic use cases, that can give you an overall idea how to use it and in which cases it might be useful.
```
VanDavv, line 4: The index page is not a tutorial, you can move this header to tutorial page

VanDavv (changes requested):
Just one thing left, otherwise looks great!

PR #661, opened 2022-03-17
Added 'debug' command line option

themarpe:
Added '--debug' option to connect to already booted devices and skips version check
CC: @szabi-luxonis @alex-luxonis as ping to use it if it comes handy


VanDavv (approved):
Thanks! One improvement we could also add is to set `DEPTHAI_LEVEL=debug` automatically once `--debug` flag is present

themarpe:
> Thanks! One improvement we could also add is to set DEPTHAI_LEVEL=debug automatically once --debug flag is present

Not possible to do cleanly at the moment:/ Will add to core first and add this in later

PR #665, opened 2022-03-21
Add temperature shutdown

VanDavv:
This PR adds a temperature shutdown mechanism that raises Overheat error if the device gets over 105* celsius

### DIFF
--- a/depthai_demo.py
+++ b/depthai_demo.py
@@ -374,7 +374,11 @@ class Demo:
         if msg.level == dai.LogLevel.CRITICAL:
             print(f"[CRITICAL] [{msg.time.get()}] {msg.payload}", file=sys.stderr)
             sys.stderr.flush()
-            self.error = OverheatError(msg.payload)
+            temperature = self._device.getChipTemperature()
+            if temperature.average > 100:
+                self.error = OverheatError(msg.payload)
+            else:
+                self.error = RuntimeError(msg.payload)
 
     timer = time.monotonic()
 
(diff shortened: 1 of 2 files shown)


depthai_demo.py
```diff
@@ -374,7 +374,11 @@ def _logMonitorCallback(self, msg: dai.LogMessage):
         if msg.level == dai.LogLevel.CRITICAL:
             print(f"[CRITICAL] [{msg.time.get()}] {msg.payload}", file=sys.stderr)
             sys.stderr.flush()
-            self.error = OverheatError(msg.payload)
+            temperature = self._device.getChipTemperature()
+            if temperature.average > 100:
```
themarpe, line 378: Modify to any temp reading > 105.0C

themarpe:
I assume you've raised the OverheatError specifically so that Sentry will pick it up as is, correct?

VanDavv:
Yes, also custom exception can be caught by users who extend the demo further (not to raise a regular RuntimeError when we can distinguish this error)

depthai_demo.py
```diff
@@ -360,6 +367,19 @@ def stop(self, *args, **kwargs):
...
 
+    def canRun(self):
+        return hasattr(self, "_device") and not self._device.isClosed()
+
+    def _logMonitorCallback(self, msg: dai.LogMessage):
+        if msg.level == dai.LogLevel.CRITICAL:
+            print(f"[CRITICAL] [{msg.time.get()}] {msg.payload}", file=sys.stderr)
+            sys.stderr.flush()
+            temperature = self._device.getChipTemperature()
+            if temperature.average > 105:
```
themarpe, line 378: Check if any temperatures is higher than 105C, otherwise it might trigger too late

themarpe (changes requested):

themarpe (approved):

PR #670, opened 2022-03-24
Added launcher fix for multiple tags on same commit

themarpe:
Fixes: https://discord.com/channels/790680891252932659/924798902627074109/956181560892194826


VanDavv (approved):
LGTM, thanks! The issue itself is surprising, as there is a possibility we could've continued along without even noticing it.

PR #673, opened 2022-03-28
Added timer for capture. 

saching13:
Default timer after space is pressed is 5 seconds. Which can be changed using the argument `-cd <number of seconds>`

Ex: `python3 calibrate.py -s 3 -brd bw1098obc -db -cd 2`

### DIFF
--- a/calibrate.py
+++ b/calibrate.py
@@ -122,6 +122,7 @@ class Main:
     images_captured = 0
 
     def __init__(self):
+        global debug
         self.args = parse_args()
         debug = self.args.debug
         self.aruco_dictionary = cv2.aruco.Dictionary_get(
--- a/calibrate.py
+++ b/calibrate.py
@@ -97,6 +97,7 @@ def parse_args():
                         required=False, help="Set capture FPS for all cameras. Default: %(default)s")
     parser.add_argument("-cd", "--captureDelay", default=5, type=int,
                         required=False, help="Choose how much delay to add between pressing the key and capturing the image. Default: %(default)s")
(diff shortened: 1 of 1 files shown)


saching13:
cc: @alex-luxonis  @daxoft  @Erol444 on testing

calibrate.py
```diff
@@ -92,9 +93,11 @@ def parse_args():
                         required=False, help="Choose between perspective and Fisheye")
     parser.add_argument("-rlp", "--rgbLensPosition", default=135, type=int,
                         required=False, help="Set the manual lens position of the camera for calibration")
-    parser.add_argument("-fps", "--fps", default=30, type=int,
+    parser.add_argument("-fps", "--fps", default=10, type=int,
                         required=False, help="Set capture FPS for all cameras. Default: %(default)s")
-    
+    parser.add_argument("-cd", "--captureDelay", default=5, type=int,
```
szabi-luxonis, line 98: Should we keep the default behaviour, thus setting this to 0? Or the intention is to default to 5 from now on?
Also defaulting to 10 fps is meant to be all cases from now on? I think it's possible that some new sensors will not support FPS changing out of the box (OAK-D lite when got mainlined), so it might cause issues if we rely on exact fps set.

Otherwise rest of the changes LGTM

saching13: > the intention is to default to 5 from now on?

Yes.  The reason being to be able to ahve time to hold the board properly. 

> I think it's possible that some new sensors will not support FPS changing out of the box

For that we will ahve a FPS setting override  like OAK-D-Lite was. 
And this is temporary. Daniel worked on a UI one which is pending now.

szabi-luxonis (approved):

PR #678, opened 2022-04-04
Added choice for video type when encoding

ZigiMigi:
Added a choice for the user to pick which file type should the encoder record to (.h264, .h265, .mjpeg and .mcap)

### DIFF
--- a/depthai_sdk/src/depthai_sdk/managers/encoding_manager.py
+++ b/depthai_sdk/src/depthai_sdk/managers/encoding_manager.py
@@ -13,9 +13,9 @@ class EncodingManager:
     _encodingQueues = {}
     _encodingNodes = {}
     _encodingFiles = {}
-    _sufix = ""
+    _mcap = {}
 
-    def __init__(self, sufix, encodeConfig: dict, encodeOutput=None):
+    def __init__(self, encodeConfig: dict, encodeOutput=None):
         """
         Args:
             encodeConfig (dict): Encoding config consisting of keys as preview names and values being the encoding FPS
@@ -23,7 +23,6 @@ class EncodingManager:
         """
         self.encodeConfig = encodeConfig
         self.encodeOutput = Path(encodeOutput) or Path(__file__).parent
-        self._sufix = sufix
 
     def createEncoders(self, pm):
         """
@@ -34,8 +33,8 @@ class EncodingManager:
         """
 
         self._encodingNodes.clear()
-        for cameraName, encFps in self.encodeConfig.items():
-            pm.createEncoder(cameraName, encFps)
+        for cameraName, encFpsSufix in self.encodeConfig.items():
+            pm.createEncoder(cameraName, encFpsSufix[0])
             self._encodingNodes[cameraName] = getattr(pm.nodes, cameraName + "Enc")
 
     def createDefaultQueues(self, device):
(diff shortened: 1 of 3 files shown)


depthai_sdk/src/depthai_sdk/managers/pipeline_manager.py
```diff
@@ -536,6 +536,15 @@ def createEncoder(self, cameraName, encFps=30, encQuality=100):
                 raise RuntimeError("Right mono camera not initialized. Call createRightCam(res, fps) first!")
             encIn = self.nodes.monoRight.out
 
+        # set preferred profile for camera if set
+        if preferredProfile:
+            if preferredProfile == ".h264":
```
VanDavv, line 541: preferredProfile should be already an instance of `dai.VideoEncoderProperties.Profile`. A correct parsing for CLI args (that are specific for demo) should be added to `ConfigManager` (which is not part of the SDK)

depthai_sdk/src/depthai_sdk/managers/encoding_manager.py
```diff
@@ -48,16 +50,24 @@ def createDefaultQueues(self, device):
...
 
     def parseQueues(self):
         """
         Parse the output queues, consuming the available data packets in it and storing them inside opened stream files
         """
         for name, queue in self._encodingQueues.items():
             while queue.has():
-                queue.get().getData().tofile(self._encodingFiles[name])
+                if self.encodeConfig[name][1] == ".mcap":
+                    self._mcap[name].imageSave(queue.get().getData(), name)
```
VanDavv, line 67: this should be a `value` of `EncodingTypes` enum (which is a class currently). 
It should behave the same as `Previews` enum - where we use `Preview.color.name` to get a stream name and `Preview.color.value(packet, <other_args>)` to get an actual frame.

Same should apply here, each encoding type will have it's own method for handling received packet and save it to the disk

depthai_sdk/src/depthai_sdk/managers/encoding_manager.py
```diff
@@ -79,8 +89,13 @@ def printManual():
         for queue in self._encodingQueues.values():
             queue.close()
 
+        for name, file in self._mcap.items():
```
VanDavv, line 92: DepthAiMcap class should have a `close` method that will allow to close everything - then, it could be used inside `self._encodingFiles` parameter and closed the same way

depthai_sdk/src/depthai_sdk/managers/encoding_manager.py
```diff
@@ -48,16 +50,24 @@ def createDefaultQueues(self, device):
 
         for cameraName, node in self._encodingNodes.items():
             self._encodingQueues[cameraName] = device.getOutputQueue(cameraName + "EncXout", maxSize=30, blocking=True)
-            suffix = ".h265" if node.getProfile() == dai.VideoEncoderProperties.Profile.H265_MAIN else ".h264"
-            self._encodingFiles[cameraName] = (self.encodeOutput / cameraName).with_suffix(suffix).open('wb')
+            # suffix = ".h265" if node.getProfile() == dai.VideoEncoderProperties.Profile.H265_MAIN else ".h264"
+            if self.encodeConfig[cameraName][1] == ".mcap":
```
VanDavv, line 54: Please use EncodingType enum for comparison

depthai_sdk/src/depthai_sdk/managers/encoding_manager.py
```diff
@@ -48,16 +50,24 @@ def createDefaultQueues(self, device):
...
+                self._mcap[cameraName] = DepthAiMcap(cameraName)
```
VanDavv, line 55: DepthAiMcap should call `imageInit` during initialization as it requires the same parameters. After adding `close` method to it, you'll be able to use `_encodingFiles` to store an object of this class

VanDavv (changes requested):

PR #682, opened 2022-04-14
Made some tests for sdk (to be run after every release, to check if everything works)

ZigiMigi:
Made tests for classes in SDK. Tests are meant to be run after every release, to find potential errors with arguments, functions, etc.
These are not all the tests, some more must be added.


ZigiMigi:
Here is a quick "intro"
These tests were made for testing DepthAi SDK, when a new version is released (some arguments might not work together, or some might have gotten removed, etc.) -> these tests will test most functions (some can not be tested...) with different arguments.
Currently made tests:
- blob manager,
- pipeline manager (main functions),
- encoding manager,
- nnet manager (test program that uses all the main functions),
- preview manager (test program that uses all the main functions),
- utils.
All other "managers" like FPS and Preview will have to be tested manually (most managers use them, so that can also be a good indicator if they work properly).
A lot of managers also change something on the frames, which can not be tested automatically ...

ZigiMigi:
ALSO, some error's were found while testing:

Pipeline manager:
- when using createLeftCam and createRightCam, the "xout" attribute has no effect (it creates a node anyway),

Encoding manager:
- in init "Path" is not needed, yet when the attribute is not given, we get an error,
- init does not check if Path is valid

Blob manager:
- when using init, giving a false path or name (zoo name, config path, zooDir, etc.) the manager prompts no error

Preview manager:
- when using Previews.depth.name in init an error occurs "RuntimeError: Depth config has to be provided before decoding depth data", this worked before ...

VanDavv:
@ZigiMigi I made a small refactor of this PR to allow easier suite execution and have a clearer split between unit tests and end to end tests. Also, while doing so, I fixed most of the issues you mentioned above. Will close this PR, but your commits will still be merged from this PR - https://github.com/luxonis/depthai/pull/686. 
Thanks a lot!

ZigiMigi:
Perfect!!

PR #686, opened 2022-04-15
Add tests to DepthAI SDK

VanDavv:
There are two types of tests available in this PR (built from https://github.com/luxonis/depthai/pull/682 from @ZigiMigi)

- Manual - these tests need to be run separately, as they contain example applications that allow us to perform end-to-end tests. Currently, we have
    - `depthai_sdk/src/test/manual/preview_manager.py` - to test out preview streams
    - `depthai_sdk/src/test/manual/nnet_manager.py` - to test out nnet inference

- Unittests - these tests can be run in a suite and are usually quick to execute. They test out specific methods and their outcome, without testing the whole flow and inter-connections. They can be executed by running the following command

  ```
  $ cd depthai_sdk/src/test
  $ python3 -m unittest
  ```


themarpe (approved):
LGTM

PR #687, opened 2022-04-15
Add useImageManip flag to NNetManager

VanDavv:
This PR adds the `useImageManip` flag to `createNN` function, which allows to disable creating the ImageManip before the neural network (requires the user to make sure the frames are in a correct resolution for NN)


themarpe (approved):
lgtm

PR #690, opened 2022-04-26
Add initial dynamic recalibration

szabi-luxonis:
Intrinsic parameters of the camera are unlikely to change after factory calibration, however extrinsic calibration can change, if the casing was removed, PCB bent etc.
Dynamic recalibration can correct camera rotation without having to perform a full calibration.

TODO:
-Handle different stereo max resolutions (e.g. OAK-D-Lite 480p max).
-Cleanup code
-Experiment with extra filtering to reduce epipolar error

### DIFF
--- a/dynamic_recalibration.py
+++ b/dynamic_recalibration.py
@@ -183,12 +183,17 @@ if __name__ == "__main__":
     cam_left.out.link(stereo.left)
     cam_right.out.link(stereo.right)
 
-
     stereo.syncedLeft.link(xout_left.input)
     stereo.syncedRight.link(xout_right.input)
     stereo.rectifiedLeft.link(xout_left_rect.input)
     stereo.rectifiedRight.link(xout_right_rect.input)
     stereo_img_shape = cam_left.getResolutionSize()
+
+    leftFps = cam_left.getFps()
+    rightFps = cam_right.getFps()
+
+    if leftFps != rightFps:
+        raise Exception("FPS between left and right cameras must be the same!")
 
     if rgbEnabled:
         rgbLensPosition = None
@@ -215,6 +220,11 @@ if __name__ == "__main__":
         rgb_cam.isp.link(xout_rgb_isp.input)
 
         rgb_img_shape = rgb_cam.getVideoSize()
+
+        rgbFps = rgb_cam.getFps()
+        if leftFps != rgbFps:
+            raise Exception("FPS between stereo cameras and rgb camera must be the same!")
+
 
     with device:
         device.startPipeline(pipeline)
--- a/dynamic_recalibration.py
+++ b/dynamic_recalibration.py
@@ -8,6 +8,10 @@ import math
 import argparse
 from pathlib import Path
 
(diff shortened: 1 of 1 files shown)


Luxonis-Brandon:
Sweet!  This will be so useful!

szabi-luxonis:
Here is a demonstration, takes only a few seconds:

https://user-images.githubusercontent.com/60790666/165219875-011155aa-0dcb-48f1-b8d1-0431f4c234a6.mp4

dynamic_recalibration.py
```diff
@@ -0,0 +1,352 @@
...
+    pts2 = []
+    for i,(m,n) in enumerate(matches):
+        if m.distance < 0.8*n.distance:
+            pts2.append(kp2[m.trainIdx].pt)
+            pts1.append(kp1[m.queryIdx].pt)
+
+    pts1 = np.float32(pts1)
+    pts2 = np.float32(pts2)
+    # this is just to get inliers
+    M, mask = cv2.findHomography(pts1, pts2, cv2.RANSAC,5.0) #TODO try and compare findHomography against recoverPose and extract rotation from obtained homography
```
edgarriba, line 100: i suggest to use `USAC` -> https://opencv.org/evaluating-opencvs-new-ransacs/

dynamic_recalibration.py
```diff
@@ -0,0 +1,352 @@
...
+    points, R_est, t_est, mask_pose = cv2.recoverPose(E, pts1,pts2, mask=mask)
+    # print(frame1.shape[::-1])
+    flags = 0
+    # flags=cv2.CALIB_ZERO_DISPARITY
+    R1, R2, P1, P2, Q, roi_left, roi_right = cv2.stereoRectify(k1, d1, k2, d2, frame2.shape[::-1], R_est, t_est, flags=flags)
+
+    return R_est, t_est, R1, R2, P1, P2, Q
+
+def calculate_epipolar_error(frame1, frame2):
+    sift = cv2.SIFT_create()
```
edgarriba, line 81: I suggest also to explore using more modern descriptors: 
- https://kornia-tutorials.readthedocs.io/en/latest/image_matching.html
- https://colab.research.google.com/github/kornia/tutorials/blob/master/source/descriptors_matching.ipynb

dynamic_recalibration.py
szabi-luxonis: Thanks, I switched to `USAC_MAGSAC` for the essential matrix too.

dynamic_recalibration.py
szabi-luxonis: I will check it out.
The current implementation is already achieving subpixel epipolar error.
One issue from UX point of view is that kornia depends on pytorch, which might not be available on all platforms, and also quite large dependency.
WIll likely add it as an option though.

edgarriba: fair enough :)

themarpe (approved):
If no bigger blockers, lets merge:)

One thing - for EEPROMv7: readCalibration2, then modify and flashCalibration back

PR #707, opened 2022-05-31
Eeprom v7 flash update in calibration

saching13:


calibrate.py
```diff
@@ -540,8 +540,9 @@ def calibrate(self):
                 left = dai.CameraBoardSocket.RIGHT
                 right = dai.CameraBoardSocket.LEFT
 
-            calibration_handler = dai.CalibrationHandler()
-            calibration_handler.setBoardInfo(self.board_config['board_config']['name'], self.board_config['board_config']['revision'])
+            calibration_handler = self.device.readCalibration2()
```
szabi-luxonis, line 543: Bump depthai to 2.15.5.0 in requirements

szabi-luxonis (changes requested):

calibrate.py
themarpe, line 543: readCalibration2 throws if EEPROM is empty. Make sure to catch / notify / handle that case.
Also, make sure to flash both USER & FACTORY sections and compare those to be equal

themarpe (changes requested):

Erol444:
@saching13 I also created a branch for that; https://github.com/luxonis/depthai/pull/708 Maybe we could use that, as it will flash name/revision only if it's empty? It was also [tested by one user](https://discord.com/channels/790680891252932659/924798475030392922/980828130858127421).

saching13:
> @saching13 I also created a branch for that; #708 Maybe we could use that, as it will flash name/revision only if it's empty? It was also [tested by one user](https://discord.com/channels/790680891252932659/924798475030392922/980828130858127421).

Sounds good. @Erol444  do you want to make that PR or should I import them here and wrap it up ?

themarpe (commented):
Is the "factory flashing" limited to factory devices? Or do you intend to user expose it? @saching13

themarpe (changes requested):
`readCalibration2` could be used instead of having to cross check "default" calibration returned by `readCalibration` against it being empty.

Otherwise lgtm

themarpe (approved):

PR #721, opened 2022-06-16
OAK-D-W board json file

saching13:


resources/boards/OAK-D-PRO-W.json
```diff
@@ -5,7 +5,7 @@
         "revision": "R3M1E3",
         "swap_left_and_right_cameras": true,
         "left_fov_deg": 127.0,
-        "rgb_fov_deg": 127.0,
+        "rgb_fov_deg": 108.0,
```
szabi-luxonis, line 8: Querying fov from intrinsic vs specs will be significantly different.
I think this is expected, and intended. Just mentioning.

saching13: True. But also good to have Proper full FOv value there in case needed for future.

szabi-luxonis (approved):

PR #722, opened 2022-06-17
Modified charuco board so it's 2.2cm per square on 15" screen

Erol444:
image


saching13:
got it. Can you rename the file to include the number of rows and columns?

saching13 (approved):

PR #726, opened 2022-06-29
enable IMU

daxoft:


themarpe (approved):
LGTM if tested

For IMU we have to expose RPC (todo in the future @szabi-luxonis )

saching13 (changes requested):
1. Add the option to Print the IMU values after the IMU update. 
2. Read those values from the queue and add a pass if values are coming in.

PR #738, opened 2022-07-12
Add oak-1 interface and jsons

daxoft:
Add oak-1 interface and jsons


themarpe (changes requested):
EEPROM contents LGTM

test_ui.py modifications are to be modified. As gui_ux branch already addresses that, I'd just modify that the `test_type` is an enumeration, which then gets checked in necessary places. And you set it once from product name (for now, this has and will be refactored by gui_ux branch)

PR #744, opened 2022-07-16
Updated ps1 script

Erol444:


themarpe (changes requested):
I think it'll be better to use the powershell script instead of bat (as shared in second example)

themarpe (approved):
LGTM if tested

Erol444:
@themarpe It now starts the actual app (so no problem with shortcut), but I get the error below - looks like depthai-demo requirements aren't being installed.
```
no changes added to commit (use "git add" and/or "git commit -a")
Current commit: a6ea0e08c33890fde25d3ab51f6e6f2f48104bd1
Current tag: None
Available DepthAI versions: ['v0.3.0.0', 'v0.4.0.0', 'v0.4.1.1', 'v1.0.0', 'v1.0.0.0', 'v2.0.0', 'v2.10.0.0', 'v2.7.2.0', 'v2.8.0.0', 'v2.9.0.0', 'v3.0.0', 'v3.0.1', 'v3.0.10', 'v3.0.11', 'v3.0.12', 'v3.0.2', 'v3.0.3', 'v3.0.4', 'v3.0.5', 'v3.0.6', 'v3.0.7', 'v3.0.8', 'v3.0.9']
Message Box (Update Available): Version 3.0.12 is available.
Current version is Unknown
Update?
Should update? False
  File "C:\Users\Erik\AppData\Local\Programs\DepthAI\depthai\depthai_demo.py", line 24, in <module>
    from depthai_sdk.managers import ArgsManager, getMonoResolution, getRgbResolution
  File "C:\Users\Erik\AppData\Local\Programs\DepthAI\depthai\depthai_sdk\src\depthai_sdk\__init__.py", line 1, in <module>
    from .fps import *
  File "C:\Users\Erik\AppData\Local\Programs\DepthAI\depthai\depthai_sdk\src\depthai_sdk\fps.py", line 3, in <module>
    import cv2
ModuleNotFoundError: No module named 'cv2'
DepthAI Demo ret code: 1
```

Erol444:
Also it prompts me to update version, and if I click `Yes` it just pops up the same `Update Available` window. Probably because this installer is not release version (has version `Unknown`, but something to try out after release.
image

moratom:
The issue with not installing dependencies that you mentioned @Erol444 seems to be an issue in the v3.0.12 release as well, at least on my machine. Looking into it now.

depthai_demo.py
```diff
@@ -21,11 +21,22 @@
...
 sys.path.append(str((Path(__file__).parent / "depthai_sdk" / "src").absolute()))
-from depthai_sdk.managers import ArgsManager
+
+
+try:
+    from depthai_sdk.managers import ArgsManager
+    import cv2
+    import depthai as dai
+    import numpy as np
+except Exception as ex:
```
Erol444, line 31: @moratom shouldn't these be installed by the installer though?

moratom:
OK, I managed to track down what the issue with dependencies is.
Basically, the installer doesn't install the `requirements.txt` since at least version 2.0.0 (on my machine that is running Windows 11). 
It was always taken care of on the first run, where this [error handling](https://github.com/luxonis/depthai/blob/92c66ab3e868915623329e703f154e6afd0b35f0/launcher/launcher.py#L268) would take care of the dependencies.
This got broken with this [commit](https://github.com/luxonis/depthai/commit/46a0461bec41a146fc4d489d580d9744745a1db9), because imports were introduced without "try and except" guards.

My commit above fixes the dependency issue, but this is completely unrelated to the shortcut issue. Should the commit stay here, or should I open a new issue?

CC: @themarpe @Erol444

depthai_demo.py
moratom: Yes, they should. But at least on my machine, the installer doesn't install them at least since version 2.0.0, while this fallback got broken more recently. I will look into why the installer doesn't install them now.

Erol444:
@moratom great! Can I test this commit locally? Just curious, where in that commit were imports introduced without "try and except" guards?

moratom:
> @moratom great! Can I test this commit locally? Just curious, where in that commit were imports introduced without "try and except" guards?

@Erol444 the import is here https://github.com/luxonis/depthai/blob/46a0461bec41a146fc4d489d580d9744745a1db9/depthai_demo.py#L13

And yes, you should be able to test this locally, just checkout this branch.

moratom:
I created a separate PR #746 for this requirements issue and removed the commit from here. As far as I can tell, this one can be closed @themarpe @Erol444 as the shortcut is working well.

themarpe (approved):

PR #746, opened 2022-07-18
Catch if missing requirements

moratom:
## Catch if there are missing requirements.

Issue found while testing PR #744.

- On my machine (running Windows 11) the installer doesn't install the requirements.txt since at least version 2.0.0.

- It was always taken care of on the first run, where this [error handling](https://github.com/luxonis/depthai/blob/92c66ab3e868915623329e703f154e6afd0b35f0/launcher/launcher.py#L268) would take care of the dependencies.

- This got broken with this https://github.com/luxonis/depthai/blob/e6d7feaa51e521b7b8a697bc2ec86caf5e1cb3b6/depthai_demo.py#L24 as imports were introduced without "try and except" guards.

This PR should solve the issue.
CC: @Erol444 @themarpe


themarpe (approved):
LGTM, Thanks!

Issue #750, opened 2022-07-21
[BUG] NameError: name ''App is not defined


spider-x:
When i press in the depthai gui button for running oak-d lite as webcam it shows me error and camera is showing as webcam.
When i run command python depthai_demo.py --app uvc it is working properly.
![image](https://user-images.githubusercontent.com/413210/180140099-d5926fdf-1123-4ed3-8f30-e236ca505dd5.png)

Erol444:
Hi @spider-x ,
Have you installed requirements `python install_requirements.py` before starting the depthai demo app?
Thanks, Erik

spider-x:
> Hi @spider-x , Have you installed requirements `python install_requirements.py` before starting the depthai demo app? Thanks, Erik

No, i have used windows installer to install the depthai. I would like to ask also if is there another command than python depthai_demo.py --app uvc to run the camera as webcam. I'm asking because, everytime i run this command, it downloads and install some things and it is time-consuming.  Tahnk you for you help.

moratom:
Hi @spider-x @Erol444. I am able to reproduce the bug - it exists both on Windows and Linux. 
Looking into it now.

moratom:
The issue should be fixed in the #752 PR @spider-x @Erol444.

PR #752, opened 2022-07-21
Include App in all cases

moratom:
Solves the #750 issue.

### DIFF
--- a/depthai_demo.py
+++ b/depthai_demo.py
@@ -30,7 +30,7 @@ try:
     from depthai_helpers.app_manager import App
 except Exception as ex:
     print("Third party libraries failed to import: {}".format(ex))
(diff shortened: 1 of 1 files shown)


depthai_demo.py
```diff
@@ -27,6 +27,7 @@
     import depthai as dai
     import numpy as np
     from depthai_sdk.managers import ArgsManager, getMonoResolution, getRgbResolution
+    from depthai_helpers.app_manager import App
 except Exception as ex:
     print("Third party libraries failed to import: {}".format(ex))
     print("Run \"python3 install_requirements.py\" to install dependencies or visit our installation page for more details - https://docs.luxonis.com/projects/api/en/latest/install/")
```
Erol444, line 33: `or visit our installation page for more details - https://docs.luxonis.com/projects/api/en/latest/install/"` - This page is only for depthai-python install (API), not depthai-demo requirements, so I would remove that text

Erol444 (approved):

PR #755, opened 2022-07-22
Implement manual focus with record app

ahmadchalhoub:
This is to allow for using manual focus when running the 'record' app of the depthai_demo.py script.
This probably needs a lot more work for proper implementation, but I just though I'd open a pull request in case it is helpful.

I would appreciate any feedback or comments :)

A sample command to run this: 
`$ python3 depthai_demo.py -app record --save color -p ~/test_output -mf 200`


Luxonis-Brandon:
Thanks for doing so!

PR #762, opened 2022-08-04
[Calibration] Graceful handling when frames cannot be retrieved

themarpe:
Main change is:
```
if recent_left is None or recent_right is None or (recent_color is None and not self.args.disableRgb):
       print("Continuing...")
```

```
if recent_left is None or recent_right is None or (recent_color is None and not self.args.disableRgb):
                if time.time() - last_frame_time > 5:
                    if self.args.disableRgb:
                        print("Error: Couldn't retrieve left and right frames for more than 5 seconds. Exiting...")
                    else:
                        print("Error: Couldn't retrieve left, rigth and color frames for more than 5 seconds. Exiting...")
                    raise SystemExit(1)
                cv2.waitKey(1)
                continue
```

Otherwise the terminal just keeps spamming with `Continuing...`

### DIFF
--- a/calibrate.py
+++ b/calibrate.py
@@ -39,13 +39,13 @@ def parse_args():
 
     Image capture requires the use of a printed OpenCV charuco calibration target applied to a flat surface(ex: sturdy cardboard).
     Default board size used in this script is 22x16. However you can send a customized one too.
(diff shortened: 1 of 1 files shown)


saching13 (approved):
LGTM

PR #786, opened 2022-09-04
Removed unsupported resolutions

Erol444:


Erol444:
Already supported in latest depthai, my bad.

PR #801, opened 2022-09-27
Daxoft/ffc 4p

daxoft:
Add FFC-4P testing, only flashing

### DIFF
--- a/batch/eeprom/DD2090_R4M1E4_oak_d_ffc_4p.json
+++ b/batch/eeprom/DD2090_R4M1E4_oak_d_ffc_4p.json
@@ -7,6 +7,6 @@
     "productName": "OAK-D FFC-4P",
     "boardCustom": "",
     "hardwareConf": "F0-FV00-BC000",
-    "boardOptions": 3,
+    "boardOptions": 0,
     "version": 7
 }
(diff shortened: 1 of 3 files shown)


batch/eeprom/DD2090_R4M1E4_oak_d_ffc_4p.json
```diff
@@ -0,0 +1,12 @@
+{
+    "batchName": "",
+    "batchTime": 0,
+    "boardConf": "IR-C00M00-00",
+    "boardName": "DD2090",
+    "boardRev": "R4M1E4",
+    "productName": "OAK-D FFC-4P",
+    "boardCustom": "",
+    "hardwareConf": "F0-FV00-BC000",
+    "boardOptions": 3,
```
themarpe, line 10: boardOptions: 0

for USB based devices. (no modifications for PCIe clock and no modifications for USB3 phy clock)

themarpe (changes requested):
Do you use getCameraNames and dynamically build output for all cameras? We could have a 4 pieces of eg 378 or some other sensor to go along to test out all the connectors

Luxonis-David (approved):

batch/eeprom/DD2090_R4M1E4_oak_d_ffc_4p.json
daxoft: @themarpe updated boardOptions, I don't show the camera streams for FFC-4P at all at @Luxonis-David's suggestion (it was also easier to implement), I could add that option (to search for cameras and show the stream when available without dropping the test if not) in a later PR

themarpe (approved):

PR #809, opened 2022-10-11
swap lr board settings used by calibrate

georgeman93:
I had some confusion whether why of the board setting (swap LR) wasn't been used, I figure its clearer to use it and explain in the help string that the `-slr` option overrides it.


saching13:
Thanks for the PR. The variable in json file is deprecated and there might be copy paste wrong values in some of the files. Using that for some files might break things. 
We have a refactor of calibration merging in 2 weeks. So I would just leave this as it is.

PR #815, opened 2022-10-19
Packet ids

daniilpastukhov:
- Add unique id for every component's output
- Add generic component output `components.component.ComponentOutput`
- Add component's id to packet's `name`


Erol444:
The problem with having component it be the unique id for each output is that you can't sync multiple outputs from the same component, eg;
```
oak.sync([color.out.encoded, nn.out.main, nn2.out.main], cb) # Works

oak.sync([color.out.encoded, nn.out.main, nn.out.passthrough, nn2.out.main], cb) # Doesn't work
```

Erol444:
Fixed here: https://github.com/luxonis/depthai/commit/8d900c77c1676648db7b0cf7a6912d55b54f9bfd

PR #819, opened 2022-10-26
Sdk complete refactor fixes

daniilpastukhov:
- Auto thickness adjustment
- Fix age-gender demo
- Replace kwargs with named arguments for visualizer's config methods

### DIFF
--- a/depthai_sdk/src/depthai_sdk/visualize/visualizer.py
+++ b/depthai_sdk/src/depthai_sdk/visualize/visualizer.py
@@ -9,7 +9,7 @@ import depthai as dai
 import numpy as np
 from depthai import ImgDetection
 
-from depthai_sdk.visualize.configs import VisConfig, TextPosition
+from depthai_sdk.visualize.configs import VisConfig, TextPosition, BboxStyle
 from depthai_sdk.visualize.encoder import JSONEncoder
 from depthai_sdk.visualize.objects import VisDetections, GenericObject, VisText, VisTrail, VisCircle
 from depthai_sdk.visualize.visualizer_helper import VisualizerHelper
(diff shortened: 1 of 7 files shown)


depthai_sdk/src/depthai_sdk/visualize/visualizer.py
```diff
@@ -199,55 +200,129 @@ def serialize(self) -> str:
...
+            img_scale: Scale of the output image.
+            show_fps: Flag that indicates if the FPS should be shown.
 
         Returns:
             self
         """
-        self.config.output = replace(self.config.output, **kwargs)
+        kwargs = locals()
+        kwargs.pop('self')
+        kwargs = {k: v for k, v in kwargs.items() if v is not None}
```
Erol444, line 221: Maybe have a function for this logic, so it's not copy pasted across all public functions?

depthai_sdk/src/depthai_sdk/visualize/visualizer.py
```diff
@@ -199,55 +200,129 @@ def serialize(self) -> str:
...
+
+        if len(kwargs) > 0:
+            self.config.text = replace(self.config.text, **kwargs)
+
         return self
 
-    def tracking(self, **kwargs: dict) -> 'Visualizer':
+    def tracking(self,
```
Erol444, line 302: I think `max_age` would also be very useful here. Or maybe `max_lenght`, so after X pixels or % of the image it would discard older tracklets. Thoughts?

Erol444 (approved):

depthai_sdk/src/depthai_sdk/visualize/visualizer.py
daniilpastukhov: Yeah, it's a good idea. Can be really useful for fast moving objects.

daniilpastukhov: Yeah, great idea. Added it in 907040bdd5636dfd5153eaf1d4e618ca372b4435. It's customizable via `visualizer.tracking(max_length=100)`, where the units are pixels. So here the maximum length cannot exceed 100px.

daniilpastukhov:
@Erol444 can you please merge if everything is fine now? Thanks.

Erol444 (approved):
Awesome, thanks!

PR #820, opened 2022-10-27
Update nn_component.

peskaf:
Update of nn_component so that multistage pipeline is configurable.

Example of code that would not run as expected with the code we currently have:
```python
from depthai_sdk import OakCamera, AspectRatioResizeMode
from depthai_sdk.components.nn_component import NNComponent

with OakCamera() as oak:
    color = oak.create_camera('color')

    det: NNComponent = oak.create_nn('face-detection-retail-0004', color)
    det.config_nn(aspectRatioResizeMode=AspectRatioResizeMode.CROP)
    emotion_nn: NNComponent = oak.create_nn('emotions-recognition-retail-0003', input=det)
    emotion_nn.config_multistage_nn(debug=True)
    
    oak.callback(emotion_nn, lambda packet: print(packet))

    oak.start(blocking=True)
```


Erol444 (approved):
Thanks for the fix!

PR #838, opened 2022-11-09
Upload results to the production support server

njezersek:
Added a mechanism to upload the test results to the stats server (specified with environment variables `PRODUCTION_SUPPORT_SERVER_URL` and `PRODUCTION_SUPPORT_SERVER_API_KEY`) and cache them in case of an outage in JSONL formatted file (specified with `CACHE_FILE_NAME` and `CACHE_DIRECTORY`).


depthai_helpers/production_support_server_api.py
```diff
@@ -61,7 +61,7 @@ def sync():
 			results = [result for result in reader]
 		
 		response = requests.post(
-			f'{STATS_SERVER_URL}/results', 
+			f'{PRODUCTION_SUPPORT_SERVER_URL}/results',
```
PetrNovota, line 64: I would love to make this more general, would be nice to be able to select table, so I would change it to something like
` f'{PRODUCTION_SUPPORT_SERVER_URL}/TABLE_NAME'`

test_ui.py
```diff
@@ -281,6 +287,20 @@ def __init__(self):
         self.current_jpeg = 0
 
         self.id = self.device.getDeviceInfo().getMxId()
+        self.bootloader_version = "library"
+        try:
+            self.bootloader_version = self.device.getBootloaderVersion().toStringSemver()
+        except: pass
```
themarpe, line 293: Lets change the semantics a bit - let `bootloader_version` denote what BL version was flashed to the device (it might be that upon a next reboot, device wouldn't boot BL, this might be the case for PoE devices where we want to perform everything/most over USB for Test stage)

Also, in some cases we flash boot header - that would preferably be denoted as well (we can do just flashRead of 512B and store as is)

So:
 - which BL version was flashed (if it was flashed)
 - which type of BL (there are two, USB and NETWORK)
 - And first 512B of flash as is (raw)

themarpe (changes requested):

depthai_helpers/production_support_server_api.py
njezersek: I think multiple tables are unnecessary since you can use the `stage` field to differentiate different stages.

themarpe (approved):
LGTM if tested and ready for rollout

njezersek:
Tested with:
- OAK-D-Pro-FF
- OAK-D-Pro-W
- OAK-D-Pro-PoE-AF

PR #839, opened 2022-11-09
Installation hotfix

daniilpastukhov:


requirements.txt
```diff
@@ -11,3 +11,4 @@ opencv-contrib-python==4.4.0.46 ; platform_machine == "armv6l" or platform_machi
 pyqt5>5,<5.15.6 ; platform_machine != "armv6l" and platform_machine != "armv7l" and platform_machine != "aarch64"
 --extra-index-url https://artifacts.luxonis.com/artifactory/luxonis-python-snapshot-local/
 depthai==2.19.0.0
+Qt.py
```
Erol444, line 14: from my understanding, we already have `pyqt5` requirement for the depthai_demo (so the qt gui), why would we need another requirement?

Erol444 (approved):

requirements.txt
daniilpastukhov: Qt.py is just a wrapper that allows users to use PyQt4/5 or PySide 1/2

PR #844, opened 2022-11-12
Demo app installation changes

PetrNovota:
requirements.txt changes:
- depthai-sdk can be installed via pip
- depthai version 2.19
- pyqt5 deleted, it cannot be installed by mac silicon machines, mac silicon has to download it via homebrew. For demo installation, this was moved to the installation script

run_demo script added: shortcut to run demo_script after installation of virtual env

optional_requirements:
let pip install latest versions available for platform for these packages
- ffmpy3
- pyusb
- sentry-sdk


requirements.txt
```diff
@@ -6,9 +6,8 @@ opencv-contrib-python==4.5.4.58 ; platform_machine != "aarch64" and platform_mac
 opencv-contrib-python==4.5.1.48 ; platform_machine != "aarch64" and platform_machine != "armv6l" and platform_machine != "armv7l" and python_version != "3.10"
 opencv-python==4.4.0.46 ; platform_machine == "armv6l" or platform_machine == "armv7l"
 opencv-contrib-python==4.4.0.46 ; platform_machine == "armv6l" or platform_machine == "armv7l"
--e ./depthai_sdk
+depthai-sdk
 --extra-index-url https://artifacts.luxonis.com/artifactory/luxonis-depthai-data-local/wheels/
-pyqt5>5,<5.15.6 ; platform_machine != "armv6l" and platform_machine != "armv7l" and platform_machine != "aarch64"
```
themarpe, line 11: CC: @daniilpastukhov @Erol444

requirements-optional.txt
```diff
@@ -1,7 +1,7 @@
 open3d==0.10.0.0; platform_machine != "armv6l" and platform_machine != "armv7l" and python_version < "3.9" and platform_machine != "aarch64"
-ffmpy3==0.2.4
-pyusb==1.2.1
-sentry-sdk==1.5.1
+ffmpy3
+pyusb
+sentry-sdk
```
themarpe, line 4:
```suggestion
ffmpy3>=0.2.4
pyusb>=1.2.1
sentry-sdk>=1.5.1
```

+1 (though it'd be better to have a min working version specified IMO)

But CC: @daniilpastukhov @Erol444  
And side question, why is `pyusb` needed?

install_requirements.py
```diff
@@ -42,6 +42,10 @@
 subprocess.check_call(pip_call + ["uninstall", "depthai", "--yes"])
 subprocess.check_call(pip_package_install + ["-r", "requirements.txt"], cwd=scriptDirectory)
 
+# arm version is installed by homebrew in installation script
+if not is_pi:
+    subprocess.check_call(pip_call + ["install", "pyqt5"])
```
themarpe, line 47: Needed? pyqt5 is already specified in requirements.txt

depthai_demo.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+#!/usr/bin/env python3.10
```
themarpe, line 1: I don't think this is a right assumption to make...

themarpe (changes requested):

install_requirements.py
PetrNovota: Needed, I deleted PyQt5 from requirements. It causes problem for m1 chips because PyQt5 for mac silicon is not available in pip

depthai_demo.py
PetrNovota: agree, its a leftover from my testing, I decided to run demo via run_demo script which sources venv and calls depthai_demo.py with appropriate python

richard-xx:
Have you thought about changing the dependency package `pyside2` of `depthai_sdk` and the dependency package `pyqt5` of `depthai_demo` to the same one?

requirements.txt
PetrNovota: > Have you thought about changing the dependency package `pyside2` of `depthai_sdk` and the dependency package `pyqt5` of `depthai_demo` to the same one?

Great idea. I will change the demo demendency to pyside2

install_requirements.py
themarpe: @PetrNovota 
You can out it based on arch, but that works as well. 
Also, does is_pi then does the right thing for all the others?
Shouldn't it be `if not is_m1`?

themarpe:
@richard-xx 

Yup, would make a lot of sense - @daniilpastukhov I think given our reliance on it, it'd be best to fixate to pyqt5 as that seems to be OKayish supported accross different systems? (Still quite weak TBH, piwheels doesn't carry it, prebuilts missing, m1,...)

requirements-optional.txt
daniilpastukhov: `pyusb` is used to make a system report (in depthai_demo.py)

install_requirements.py
daniilpastukhov: I'd rename `is_pi` to `is_arm`

requirements.txt
daniilpastukhov: Don't we need pyqt5 to run depthai_demo.py? I support depthai_sdk has pyqt5 in its requirements, but we will remove them in the upcoming merges. It'd be better to keep it here IMO.

daniilpastukhov (commented):
Looks good except for few minor things, I've left comments there

requirements.txt
themarpe: Yes, pyqt5 should stay (preferably here) - mostly wondered for the `-e ./depthai-sdk` change

daniilpastukhov: I think editable mode isn't necessary (e.g. in case smb is pulling the updates and doesn't want `depthai_sdk` to update). Also it's more convenient for controlling what version is installed.

install_requirements.py
PetrNovota: I agree that is_py == True when arch is arm*, that should hold true for mac silicon as well as for other arm platforms

requirements.txt
PetrNovota: I think the `-e ./depthai-sdk` is not necessary, if you think it is better the way it is, I will keep it like this. 

to pyqt5 i will add platform_machine != arm64 so that it is not installed on mac silicon

install_requirements.py
PetrNovota: ill rename it to is_arm

PetrNovota:
I did not manage to install oencv-python on mac silicon using python 3.9.9, which is my global one. It looks like opencv-python wheel for mac silicon is not available. That would mean we have to use python 3.10. Any ideas? I will have to work on tv_calibration next couple days intensivelly so I might be off. It basically works, just need to solve the question, what versions are supported, on most machines, python 3.8, 3.9 will likely work, on mac silicon, python 3.10. seems to be the only option.

themarpe:
@PetrNovota 

> I did not manage to install oencv-python on mac silicon using python 3.9.9

https://pypi.org/project/opencv-python/4.6.0.66/#files

Latest opencv-python should be available for 3.7 and up on M1 (arm64 macos)

What was the issue? Was your python running under Rosetta perhaps?

PetrNovota:
> 

@themarpe the problem is that in requirements, we force m1 mac with python < 3.10 to install:
opencv-python==4.5.1.48
and it results in pip trying to build numpy from souce which results in error. If I can change the opencv-python version to 4.6.0.66 than its fine,
So maybe we just change it. to
opencv-python>=4.5.1.48?

PetrNovota (commented):
All problems are solved

PetrNovota:
I think everything is corrected, should I create new pull request? Sorry, not sure about the workflow

daniilpastukhov:
I’ll review the PR a bit later, thanks

requirements-optional.txt
themarpe: @PetrNovota 

Actually - I misread these as being SDK deps - in case of these being end application dependencies, should rather be pinned

install_requirements.py
```diff
@@ -42,6 +42,10 @@
 subprocess.check_call(pip_call + ["uninstall", "depthai", "--yes"])
 subprocess.check_call(pip_package_install + ["-r", "requirements.txt"], cwd=scriptDirectory)
 
+# arm version is installed by homebrew in installation script
+if not is_arm:
+    subprocess.check_call(pip_call + ["install", "pyqt5"])
+
```
themarpe, line 48: `is_arm` is slightly vague here (and the comment more so)

`platform.machine()` starts with "arm" for:
 - armv6l - ex: Linux (Raspbian) - RPi 1 afaik
 - armv7l - ex Linux (Raspbian) - RPi 4
 - arm64 - ex: Darwin (macOS) - M1 

But also missing aarch64 (ex: Linux (Ubuntu 64bit) - RPi 4, Jetson AFAIK, et al)

Regardless, this shouldn't be needed anymore, as the pyqt5 IS installed by requirements and properly skipped for non platforms that don't carry wheels

install_requirements.py
```diff
@@ -27,9 +27,9 @@
 if sys.version_info[0] != 3:
     raise RuntimeError("Demo script requires Python 3 to run (detected: Python {})".format(sys.version_info[0]))
 
-is_pi = thisPlatform.startswith("arm")
+is_arm = thisPlatform.startswith("arm")
 prebuiltWheelsPythonVersion = [7,9]
-if is_pi and sys.version_info[1] not in prebuiltWheelsPythonVersion:
+if is_arm and sys.version_info[1] not in prebuiltWheelsPythonVersion:
     print("[WARNING] There are no prebuilt wheels for Python 3.{} for OpenCV, building process on this device may be long and unstable".format(sys.version_info[1]))
```
themarpe, line 33: This actually points to pi.

Modify as such:

```suggestion
# Pi is a common platform using these instruction sets
is_pi = thisPlatform == 'armv6l' or thisPlatform == 'armv7l'
if is_pi and sys.version_info[1] not in piPrebuiltWheelsPythonVersion:
    print("[WARNING] There are no prebuilt wheels for Python 3.{} for OpenCV, building process on this device may be long and unstable".format(sys.version_info[1]))
```

themarpe (changes requested):

install_requirements.py
themarpe:
```suggestion
```

requirements.txt
```diff
@@ -1,14 +1,13 @@
 requests==2.26.0
 --extra-index-url https://www.piwheels.org/simple
-opencv-python==4.5.4.58 ; platform_machine != "aarch64" and platform_machine != "armv6l" and platform_machine != "armv7l" and python_version == "3.10"
-opencv-python==4.5.1.48 ; platform_machine != "aarch64" and platform_machine != "armv6l" and platform_machine != "armv7l" and python_version != "3.10"
 opencv-contrib-python==4.5.4.58 ; platform_machine != "aarch64" and platform_machine != "armv6l" and platform_machine != "armv7l" and python_version == "3.10"
-opencv-contrib-python==4.5.1.48 ; platform_machine != "aarch64" and platform_machine != "armv6l" and platform_machine != "armv7l" and python_version != "3.10"
-opencv-python==4.4.0.46 ; platform_machine == "armv6l" or platform_machine == "armv7l"
+opencv-contrib-python==4.5.1.48 ; platform_machine != "aarch64" and platform_machine != "armv6l" and platform_machine != "armv7l" and platform_machine != "arm54" and python_version != "3.10"
```
themarpe, line 4:
```suggestion
opencv-contrib-python==4.5.1.48 ; platform_machine != "aarch64" and platform_machine != "armv6l" and platform_machine != "armv7l" and platform_machine != "arm64" and python_version != "3.10"
```

daniilpastukhov:
> @daniilpastukhov
> Might make sense opening a new PR such that the previously merged CI for testing OS Support are triggered as well

Merged them here, so I think no need for the new PR

themarpe:
CC: @szabi-luxonis on 3e043da577f27d4abaf2544013017560a5e8c246 commit - I've simplified dependencies based on available ones

```
opencv-contrib-python==4.6.0.66 ; platform_machine != "armv6l" and platform_machine != "armv7l"
opencv-contrib-python==4.5.5.62 ; platform_machine == "armv6l" or platform_machine == "armv7l"
```

I assume your change for 3.10 was done just to be sure that old behavior remained, correct?

szabi-luxonis:
> CC: @szabi-luxonis on 3e043da577f27d4abaf2544013017560a5e8c246 commit - I've simplified dependencies based on available ones
> 
> 
> 
> ```
> 
> opencv-contrib-python==4.6.0.66 ; platform_machine != "armv6l" and platform_machine != "armv7l"
> 
> opencv-contrib-python==4.5.5.62 ; platform_machine == "armv6l" or platform_machine == "armv7l"
> 
> ```
> 
> 
> 
> I assume your change for 3.10 was done just to be sure that old behavior remained, correct?

Yes

daniilpastukhov:
Checks are ok now (https://github.com/luxonis/depthai/actions/runs/3496952926), can we merge? cc @themarpe

launcher/launcher.py
```diff
@@ -4,6 +4,10 @@
 
 # Standard imports
 import os, sys, subprocess, time, threading, argparse, datetime
+print(f"\nSYSTEM CHECK\n")
+print(f"{sys.prefix = }\n{sys.base_prefix = }")
```
themarpe, line 8: Seems to not work Python<3.8. Fixed

themarpe (approved):

PR #846, opened 2022-11-14
Test install requirements

moratom:
Add a CI for testing if dependencies and requirements can be installed on the supported platforms.

### DIFF
--- a/.github/workflows/test_install_requirements.yml
+++ b/.github/workflows/test_install_requirements.yml
@@ -1,4 +1,4 @@
- name: Test requirements installation
+ name: OS Support
 
  on:
   workflow_dispatch:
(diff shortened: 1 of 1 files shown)


themarpe (commented):
LGTM after applying discussed changes locally

themarpe (approved):
Thanks!

moratom:
I added a test for Windows so the three (major) platforms are covered.
The CI for macOS fails as it tries to install  `numpy==1.19.3` version that is a dependency of `opencv-python`,
and because `numpy` wheels for `python3.11` aren't built it tries to build it from source and fails. 
This is likely also what happens on fresh installs of macOS, provided that github runner reflects actual
machines with a fresh install. CC @daniilpastukhov 

@themarpe, imo we should add the test for requirements installation for the other two platforms for `depthai-python/examples/install_requirements.py`. Thoughts?

PR #849, opened 2022-11-21
Updated `calibrate.py`

njezersek:


Erol444:
Can we merge this? @njezersek @saching13

njezersek:
OAK-D-LITE has very different focus distance [m] to lens position mapping than other cameras. Therefore I changed the default lens position for OAK-D-LITE to 25 (based on trying several different values and comparing the results). Default lens position for other cameras is still 135.

engrmusawarali71:
Hello! The current script calibrate.py still shows lens position as 135. I am trying to calibrate OAK-D-Lite but images are blurry with 135. I wanted to know what should be exact and correct value @njezersek

engrmusawarali71:
Thanks @MaticTonin, I am wondering why the images are blur if the focus range is set to >60. Is it the limitation of OAK- D-Lite?

themarpe:
@engrmusawarali71 - the range can vary between devices. Certain devices have image in focus at certain ranges of lensPositions and other at different.

engrmusawarali71:
@themarpe  Thank you, I was wondering if something is wrong with the device or there is some fault.

PR #850, opened 2022-11-21
Add python versions workflows

moratom:
Test different python versions as well. 
I didn't include versions for fedora, since `setup-python` action doesn't work out of the box
in containers in some cases, but we have the Fedora listed as supported by the community only officially,
so testing only "stock" pythons for Fedora should be fine.


themarpe (approved):
LGTM
Thanks!

PR #855, opened 2022-11-25
Adding interpolation for imu msgs

vitakomel:


Erol444:
Awesome, thanks @vitakomel ! 

A few additons:
- `sync_mode` can be enum, not string
- Add integration of `fillImuData_LinearInterpolation` into `Imu()` function. So `Imu` function would work the same as the toRosMsg()` (below) - maybe by defalt we can do interpolation, otherwise we copy gyro/accel directly (so it works exactly the same way as in depthai-ros

```
void ImuConverter::toRosMsg(std::shared_ptr<dai::IMUData> inData, std::deque<ImuMsgs::Imu>& outImuMsgs) {
    if(_syncMode != ImuSyncMethod::COPY) {
        FillImuData_LinearInterpolation(inData->packets, outImuMsgs);
    } else {
        for(int i = 0; i < inData->packets.size(); ++i) {
            auto accel = inData->packets[i].acceleroMeter;
            auto gyro = inData->packets[i].gyroscope;
            outImuMsgs.push_back(CreateUnitMessage(accel, gyro));
        }
    }
```

Erol444:
LGTM!

PR #865, opened 2022-12-07
use bootloader option from jsons

daxoft:


themarpe:
IMO lets refactor this more holistically

Also ENUMs please or specify options as constants

These should be their own object in the board file, more cases will go into it not just BL.

And 3 levels of options
  1. defaults (some default specified in test_ui.py)
  2. overriden by options specified in device
  3. overriden by options of a specific variant

Sync with @njezersek

daxoft:
applied the changes @themarpe please review

themarpe (approved):

PR #866, opened 2022-12-08
Fix exceptions that occured because of reading the EEPROM data at inappropriate time

njezersek:
The error showd when running the program with unflashed OAK-D-LITE because I was trying to read the EEPROM when none were flashed yet.

To fix the issue I added `orDefault` to `readCalibration` and `readFactoryCalibration` calls. This made the error disapear but logged an empty EEPROM. So I moved the EEPROM reading code further down the code path where in all cases the device is flashed.

Also I removed the duplicated function `flash_eeprom` with prior discussion with @alex-luxonis (see 11acce4d30ecad83958cc8d899c50a2f7b5e34fc), because at first I thought it could be a good place to store the EEPROM data for logging. But later I decided to move the logging even further down the code path. If the removal poses to much risk, we can revet the commit.

### DIFF
--- a/test_ui.py
+++ b/test_ui.py
@@ -328,9 +328,9 @@ class DepthAICamera():
         self.device_name = self.device.getDeviceName()
 
         self.eepromUnionData = {}
-        calibHandler = self.device.readCalibration2()
+        calibHandler = self.device.readCalibrationOrDefault()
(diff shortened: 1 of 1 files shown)


njezersek:
Tested with:
- OAK-D-LITE
- OAK-D-PRO-W

both had their EEPROM erased before the test.

themarpe (approved):
LGTM if tested

PR #871, opened 2022-12-09
Implemented multiple cameras check

njezersek:
If more than one camera is connected an error is shown and the test process is terminated:

image


themarpe (approved):
LGTM, thanks!

PR #877, opened 2022-12-15
Ffc replay support

Erol444:
Replay module to support OAK FFC by specifying ports (instance num)


saching13:
Note: We need to remove those AVI files

PR #902, opened 2023-01-14
Updated 180 deg rotation.

peskaf:
DDR usage for pipeline below for prev. version of 180 deg rotation: 98_141_184/357_154_815 

DDR usage for pipeline below for updated version of 180 deg rotation: 73_254_912/357_154_815

```python
from depthai_sdk import OakCamera, CameraComponent

with OakCamera(rotation=180) as oak:
    cam_component: CameraComponent = oak.create_camera(source="color", fps=10, resolution="1080p")
    oak.visualize(cam_component)
    oak.start(blocking=True)
```


daniilpastukhov (approved):
LGTM

PR #907, opened 2023-01-17
Added an option to name the streams

njezersek:
> Please note that I am not very familiar with the codebase so if there is something I have missed I am open to changing the implementation or even closing the request.

### Naming the x out streams
Right now it seems to be very hard to know from which camera (left, right, rgb, ...) a packet came from.

Take a look at this example:
```python

with OakCamera() as oak:
	color = oak.create_camera('color')
	left = oak.create_camera('left')
	right = oak.create_camera('right')
	oak.sync([color.out.main, left.out.main, right.out.main], callback=lambda p: print(p))
	oak.start(blocking=True)

# prints
# {'0_preview': <depthai_sdk.classes.packets.FramePacket object at 0x00000164DDCA3400>, '1_out': <depthai_sdk.classes.packets.FramePacket object at 0x00000164DDCA3310>, '2_out': <depthai_sdk.classes.packets.FramePacket object at 0x00000164DDCA3340>}
```

As you can see the frames are labeled with "0_preview", "1_out", "2_out" which is not very developer friendly. Ideally, I would like to have the streams named the same as the camera source ("color", "left", "right" in this example) but I am afraid this would break existing code so I'm proposing a middle ground solution (with minimal change to the SDK source) which allows the user to manually set the stream name when creating a camera.

Here is an example:
```python

with OakCamera() as oak:
	color = oak.create_camera('color', name="color")
	left = oak.create_camera('left', name="left")
	right = oak.create_camera('right', name="right")
	oak.sync([color.out.main, left.out.main, right.out.main], callback=lambda p: print(p))
	oak.start(blocking=True)

# prints
# {'right': <depthai_sdk.classes.packets.FramePacket object at 0x0000012E3FCB80D0>, 'left': <depthai_sdk.classes.packets.FramePacket object at 0x0000012E3FCB89D0>, 'color': <depthai_sdk.classes.packets.FramePacket object at 0x0000012E3FCB8550>}
```

The frames are labeled as specified in the `create_camera` function.

### Fixed `setup.py`

I have also made a small change to the `setup.py`. The `requirements.txt` contains some flags (`--extra-index-url ...`) but the `setup` function expects just a list of requirements. So I added a filter to remove the flags. Please note that you have to provide these flags manually when installing the package like so:
```
pip install --extra-index-url https://artifacts.luxonis.com/artifactory/luxonis-python-snapshot-local/ -e ./depthai_sdk
```

### DIFF
--- a/depthai_sdk/setup.py
+++ b/depthai_sdk/setup.py
@@ -1,7 +1,6 @@
 import io
 
 from setuptools import setup
-import pkg_resources
 
 with open('requirements.txt') as f:
     requirements = f.readlines()
--- a/depthai_sdk/setup.py
+++ b/depthai_sdk/setup.py
@@ -1,9 +1,13 @@
 import io
 
 from setuptools import setup
+import pkg_resources
 
 with open('requirements.txt') as f:
-    required = f.readlines()
+    requirements = f.readlines()
+
+install_requires=[requirement for requirement in requirements if '--' not in requirement]
+
 
 setup(
     name='depthai-sdk',
@@ -18,7 +22,7 @@ setup(
     license='MIT',
     packages=['depthai_sdk'],
     package_dir={"": "src"},  # https://stackoverflow.com/a/67238346/5494277
-    install_requires=required,
+    install_requires=install_requires,
     include_package_data=True,
     extras_require={
         "visualize": ['PySide2',
--- a/depthai_sdk/src/depthai_sdk/oak_camera.py
+++ b/depthai_sdk/src/depthai_sdk/oak_camera.py
@@ -104,6 +104,7 @@ class OakCamera:
                           None, str, dai.ColorCameraProperties.SensorResolution, dai.MonoCameraProperties.SensorResolution] = None,
                       fps: Optional[float] = None,
                       encode: Union[None, str, bool, dai.VideoEncoderProperties.Profile] = None,
+                      name: Optional[str] = None,
                       ) -> CameraComponent:
         """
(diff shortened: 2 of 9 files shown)


depthai_sdk/src/depthai_sdk/oak_camera.py
```diff
@@ -123,6 +124,7 @@ def create_camera(self,
                                encode=encode,
                                rotation=self._rotation,
                                replay=self.replay,
+                               name=name,
```
Erol444, line 127: Add the name to the comments underneath the method (both inside oak_camera and CameraComponent), so it gets auto-populated to the documentation. Also perhaps we could add this functionality to other components?

Erol444 (approved):

depthai_sdk/setup.py
```diff
@@ -1,9 +1,13 @@
 import io
 
 from setuptools import setup
+import pkg_resources
```
Erol444, line 4: Leave this out perhaps?

Erol444 (approved):

daniilpastukhov (changes requested):
The encoded output isn't named correctly in the following example:
```python

with OakCamera() as oak:
    color = oak.create_camera('color', name="color", encode='h264')
    oak.sync([color.out.camera, color.out.encoded], callback=lambda p: print(p))
    oak.start(blocking=True)
```
Also, two outputs from the same stream won't work since they will have the same name, perhaps we could use the name as a prefix/postfix or add distinct ids?
I think it'd be useful to add naming for NNComponent and StereoComponent as well.

daniilpastukhov (commented):
LGTM

PR #921, opened 2023-01-28
SDK optimization (1.9.3)

daniilpastukhov:
- Add support for depthai's universal OpenVINO version.
- By default, `OakCamera.callback` won't initialize visualizer. It can be enabled via setting `enable_visualizer` argument to `True`.
- Now callback functions should expect only 1 argument - `packet`. Visualizer is a part of a packet now (`packet.visualizer`).
- Split `Xout*` classes from `xout.py` into dedicated files (each located in `depthai.oak_outputs.xout` subpackage)
- Make most of dependencies (cv2, av, etc) not mandatory.
- Improve imports to facilitate usage without unused uninstalled dependencies.


depthai_sdk/examples/CameraComponent/mono_400p.py
```diff
@@ -1,10 +1,8 @@
...
 
 with OakCamera() as oak:
-    left = oak.create_camera('left', resolution='400p', fps=60)
-    right = oak.create_camera('right', resolution='400p', fps=60)
-
-    oak.visualize([left, right], fps=True)
+    left = oak.create_camera('left', resolution='400p', fps=60, encode='h264')
 
+    # oak.visualize([left, right], fps=True)
+    oak.callback(left, callback=lambda packet: print(packet.imgFrame.getSequenceNum()))
```
Erol444, line 7: Probably not intended to be pushed?

depthai_sdk/examples/NNComponent/age-gender.py
```diff
@@ -1,10 +1,14 @@
...
-from depthai_sdk import OakCamera, TextPosition, Visualizer, TwoStagePacket
+from depthai_sdk import OakCamera
+from depthai_sdk.classes import TwoStagePacket
+from depthai_sdk.visualize.configs import TextPosition
 
 
-def callback(packet: TwoStagePacket, visualizer: Visualizer):
+def callback(packet: TwoStagePacket):
+    visualizer = packet.visualizer
+    print(visualizer.serialize())
```
Erol444, line 11: is this print intended?

depthai_sdk/examples/StereoComponent/stereo_encoded.py
```diff
@@ -1,9 +1,7 @@
-import cv2
-
 from depthai_sdk import OakCamera
-from depthai_sdk.visualize.configs import StereoColor
 
 with OakCamera() as oak:
     stereo = oak.create_stereo('400p', fps=30, encode='h264')
-    oak.visualize(stereo.out.encoded)
+    # oak.visualize(stereo.out.encoded)
+    oak.callback(stereo.out.encoded, callback=lambda x: print(x))
```
Erol444, line 6: same as above

depthai_sdk/setup.py
```diff
@@ -5,12 +5,11 @@
...
     requirements = f.readlines()
 
-install_requires=[requirement for requirement in requirements if '--' not in requirement]
-
+install_requires = [requirement for requirement in requirements if '--' not in requirement]
 
 setup(
     name='depthai-sdk',
-    version='1.9.3',
+    version='1.9.2.1',
```
Erol444, line 12: 1.9.3 to 1.9.2.1?

Erol444 (approved):

depthai_sdk/setup.py
daniilpastukhov: The last released version was 1.9.2, so I think this one can be 1.9.2.1. Next bigger release would be 1.9.3 then. What do you think?

Erol444: I think we can use the 4th number for bugfixes only - so for small releases we change 3rd num, and for big ones we can `num++` 2nd one. Thoughts?

daniilpastukhov: Yeah, totally agree. Changing this to 1.9.3, next one will be 1.9.4👌

daniilpastukhov:
LGTM, thanks @Erol444

PR #922, opened 2023-01-31
Create eepromUnionData on top

moratom:
Fix for `test_ui.py` as it crashed out on FFC cameras now - `init()` returned early.


njezersek (approved):
LGTM

PR #936, opened 2023-02-08
On-device stereo colorization

daniilpastukhov:
Usage:
```python
import depthai as dai

with OakCamera() as oak:
    stereo = oak.create_stereo('800p', fps=30, encode='h264')
    stereo.set_colormap(dai.Colormap.STEREO_TURBO)
    oak.visualize(stereo.out.encoded)
    oak.start(blocking=True)

```

### DIFF
--- a/depthai_sdk/src/depthai_sdk/oak_camera.py
+++ b/depthai_sdk/src/depthai_sdk/oak_camera.py
@@ -175,8 +175,7 @@ class OakCamera:
                       left: Union[None, dai.Node.Output, CameraComponent] = None,  # Left mono camera
                       right: Union[None, dai.Node.Output, CameraComponent] = None,  # Right mono camera
                       name: Optional[str] = None,
-                      encode: Union[None, str, bool, dai.VideoEncoderProperties.Profile] = None,
-                      colormap: dai.Colormap = None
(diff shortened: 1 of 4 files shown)


depthai_sdk/src/depthai_sdk/oak_camera.py
```diff
@@ -175,7 +175,8 @@ def create_stereo(self,
                       left: Union[None, dai.Node.Output, CameraComponent] = None,  # Left mono camera
                       right: Union[None, dai.Node.Output, CameraComponent] = None,  # Right mono camera
                       name: Optional[str] = None,
-                      encode: Union[None, str, bool, dai.VideoEncoderProperties.Profile] = None
+                      encode: Union[None, str, bool, dai.VideoEncoderProperties.Profile] = None,
+                      colormap: dai.Colormap = None
```
Erol444, line 179: What if we expose the colormap setting via other api, not on init?

Erol444 (changes requested):

depthai_sdk/src/depthai_sdk/oak_camera.py
daniilpastukhov: Hm, we also expose it in `StereoComponent.config_stereo`, what do you think?

Erol444 (approved):

PR #938, opened 2023-02-09
Fix cv2 import

daniilpastukhov:
- Make `import cv2` conditional (in case user doesn't have cv2 installed, app that doesn't use it shouldn't fail)


themarpe (approved):
LGTM - but do we then also handle that cv2 == None upon usage?

daniilpastukhov:
We don't. `Previews` is primarily used in depthai_demo.py, and it yields 'cv2 not installed, please do it` error when used without cv2

PR #1013, opened 2023-05-03
Peripheral calibration merge

njezersek:


depthai_helpers/calibration_utils.py
```diff
@@ -536,44 +453,37 @@ def calibrate_stereo(self, allCorners_l, allIds_l, allCorners_r, allIds_r, imsiz
                 return -1, "Stereo Calib failed due to less common features"
 
         stereocalib_criteria = (cv2.TERM_CRITERIA_COUNT +
-                                cv2.TERM_CRITERIA_EPS, 50000, 1e-9)
+                                cv2.TERM_CRITERIA_EPS, 1000, 1e-9)
```
saching13, line 456: lets keep it high. It might take longer. but would prefer a longer optimization than okish calibration

depthai_helpers/calibration_utils.py
```diff
@@ -593,287 +503,113 @@ def calibrate_stereo(self, allCorners_l, allIds_l, allCorners_r, allIds_r, imsiz
...
+                #          (0, line_row), (img_concat.shape[1], line_row),
+                #          (0, 255, 0), 1)
+                # line_row += 30
+
+            # line_col = 0
+            # while not isHorizontal and line_col < img_concat.shape[1]:
+            #     cv2.line(img_concat,
+            #              (line_col, 0), (line_col, img_concat.shape[0]),
+            #              (0, 255, 0), 1)
+            #     line_col += 40
```
saching13, line 585: lets clean up this commented code

depthai_helpers/calibration_utils.py
```diff
@@ -11,6 +11,8 @@
 import cv2.aruco as aruco
 from pathlib import Path
 from collections import deque
+from depthai_helpers.image_scaler import ImageScaler
+
 # Creates a set of 13 polygon coordinates
 traceLevel = 2
```
saching13, line 17: lets add this trave level as a parameter to calibrate.py ?

saching13 (approved):
Overall LGTM. 
Just some small optimizations. 

And can you add additional condition such that if the FOV of the cameras in the board.json is less than 100 and a fisheye mode is enabled. show a warning about that ?

depthai_helpers/calibration_utils.py
njezersek: Done. Trace level is now a `StereoCalibration` class member and is passed from `calibrate.py` when the class is constructed.

PR #1017, opened 2023-05-11
Add option for enabling replay looping

daniilpastukhov:
- API: `Replay.set_loop(True/False)`
- Works for video only as of now.


Erol444:
@daniilpastukhov maybe we change API to just `oak.replay.loop(True)`?

Erol444 (approved):

daniilpastukhov:
I think it's more intuitive to have a verb there (like `set`) that indicates what this method does

PR #1018, opened 2023-05-11
Fix max out size for stereo

daniilpastukhov:


depthai_sdk/src/depthai_sdk/components/stereo_component.py
```diff
@@ -288,7 +293,12 @@ def set_colormap(self, colormap: dai.Colormap):
             colormap_manip = self.node.getParentPipeline().create(dai.node.ImageManip)
             colormap_manip.initialConfig.setColormap(colormap, self.node.initialConfig.getMaxDisparity())
             colormap_manip.initialConfig.setFrameType(dai.ImgFrame.Type.NV12)
-            h, w = self.left.stream_size
+            if self._align_component:
+                h, w = self._align_component.node.getVideoSize() \
```
Erol444, line 297: I think it's actually ISP not video (haven't tested tho)
```suggestion
                h, w = self._align_component.node.getIspSize() \
```

Erol444 (approved):

PR #1021, opened 2023-05-14
BoundingBox class, fix 2stage NN, added facial landmark detection NNs, removed distinctipy

Erol444:


Erol444:
Distinctipy needs more than 2sec to generate 64 colors, almost 3sec for COCO80 set. This is blocking and at startup, which is very bad. [Comparison code here](https://gist.github.com/Erol444/747d5a8b711f3f8e3ce0c08931f5c7a2). Comparison of colors:
image

PR #1022, opened 2023-05-15
2-stage NN hotfixes

daniilpastukhov:


Erol444:
@daniilpastukhov  I just pushed the fix here: https://github.com/luxonis/depthai/pull/1021/commits/0f5d0948da224fe99701d642d76ce9bb96760403

depthai_sdk/src/depthai_sdk/components/nn_component.py
```diff
@@ -146,6 +146,7 @@ def __init__(self,
         self.image_manip.inputImage.setQueueSize(2)
 
         if isinstance(self._input, CameraComponent):
+            self._change_resize_mode(ResizeMode.LETTERBOX)
```
Erol444, line 149: issue here is, that if user sets it later explicitly to eg. crop it will mess everything up (as ImageManip wasn't reset)

Erol444 (changes requested):

depthai_sdk/examples/NNComponent/object_tracking.py
```diff
@@ -7,6 +7,7 @@
     # https://docs.luxonis.com/projects/sdk/en/latest/features/ai_models/#sdk-supported-models
     nn = oak.create_nn('yolov6nr3_coco_640x352', color, tracker=True)
 
+    nn.config_nn(resize_mode='stretch')
```
Erol444, line 10: Not needed

Erol444 (approved):

depthai_sdk/src/depthai_sdk/components/nn_component.py
daniilpastukhov: Removed that

Issue #1024, opened 2023-05-18
[BUG] NNET callbacks without visualizer enabled don't work in 1.10.0


zrezke:
Depthai Sdk crashes when trying to use callbacks on nnets without the visualizer enabled.

**MRE**
```py
from depthai_sdk import OakCamera

with OakCamera() as oak:
    color = oak.create_camera("color")
    nnet = oak.create_nn("yolov8n_coco_640x352", color)
    oak.callback(nnet, lambda frame: print(frame))
    oak.start(blocking=True)
```

**Traceback**
```sh
Traceback (most recent call last):
  File "/home/filipjeretina/Documents/dev/depthai-sdk-experiments/nnet_cb.py", line 9, in <module>
    oak.start(blocking=True)
  File "/home/filipjeretina/Documents/dev/depthai-viewer/venv/lib/python3.10/site-packages/depthai_sdk/oak_camera.py", line 353, in start
    self.poll()
  File "/home/filipjeretina/Documents/dev/depthai-viewer/venv/lib/python3.10/site-packages/depthai_sdk/oak_camera.py", line 381, in poll
    self._oak.check_sync()
  File "/home/filipjeretina/Documents/dev/depthai-viewer/venv/lib/python3.10/site-packages/depthai_sdk/oak_device.py", line 44, in check_sync
    sync.check_queue(block=False)  # Don't block!
  File "/home/filipjeretina/Documents/dev/depthai-viewer/venv/lib/python3.10/site-packages/depthai_sdk/oak_outputs/xout/xout_base.py", line 91, in check_queue
    self.on_callback(packet)
  File "/home/filipjeretina/Documents/dev/depthai-viewer/venv/lib/python3.10/site-packages/depthai_sdk/oak_outputs/xout/xout_nn.py", line 102, in on_callback
    self._visualizer.frame_shape = self._frame_shape
AttributeError: 'NoneType' object has no attribute 'frame_shape'
```

PR #1025, opened 2023-05-18
Check if visualizer is enabled before trying to set frame_shape, fixes https://github.com/luxonis/depthai/issues/1024

zrezke:


Erol444 (approved):
LGTM, thanks!

PR #1039, opened 2023-05-25
Enable runtime mode switch, so you can toggle lr_check, extended and subpixel in runtime with StereoControl

zrezke:


depthai_sdk/src/depthai_sdk/components/stereo_component.py
```diff
@@ -173,7 +173,8 @@ def __init__(self,
         self._control_xlink_in.setStreamName(f"{self.node.id}_inputControl")
         self._control_xlink_in.out.link(self.node.inputConfig)
         self._control_xlink_in.setMaxDataSize(1) # CameraControl message doesn't use any additional data (only metadata)
-
+        self.node.setRuntimeModeSwitch(True)
```
Erol444, line 176: It's a bad idea to have this enabled by default, as it consumes max resources, which can result in lower FPS. We enabled it on time (long ago) in depthai_demo and the first user experience was to get 23FPS on simple object detection model (instead of 30).

Erol444 (changes requested):

depthai_sdk/src/depthai_sdk/components/stereo_component.py
zrezke: Makes sense.. Wasn't sure if it was left out on purpose or not. Thanks!

PR #1053, opened 2023-06-14
added live epipolar check for oak's

saching13:


themarpe:
CC: @Erol444 @daniilpastukhov 

Lets target this to be part of SDK

Also, needs to be generilized

PR #1057, opened 2023-06-20
Add support to cv2 with no GUI support

daniilpastukhov:


depthai_sdk/src/depthai_sdk/oak_camera.py
```diff
@@ -399,7 +400,7 @@ def poll(self) -> Optional[int]:
 
         Returns: key pressed from cv2.waitKey, or None if
         """
-        if cv2:
+        if cv2 and CV2_HAS_GUI_SUPPORT:
```
OpeningNew, line 403: CV2_HAS_GUI_SUPPORT is enough

PR #1068, opened 2023-06-28
Update launcher to let the user choose between running the Depthai Demo or Depthai Viewer

zrezke:
After the launcher updates itself, the user will be prompted to choose between running depthai demo or depthai viewer (on every startup). 

- If the viewer is chosen and not installed, it will be installed. 
- If a new viewer version is available on pypi, a dialog will pop up showing what version is installed and which new version is available. The user can then confirm to update or keep using the currently installed version.

## Choose app dialog
image

## Depthai Viewer update dialog
image


themarpe:
Lets retain old logo on demo and old blue color on it.

And add `New` `Beta` on Viewer

Also, check to see if it works on all 3 major OSs

zrezke:
### New look
image

Works on windows.
Will test on linux and macos now, it for sure won't work on arm macs without rosetta due to PyQt5, which is also used in the demo. I could port the launcher to use PyQt6 tho, should be a simple task. It may be worth it if we plan to keep the launcher even after the demo is deprecated, which we do afaik.  Thoughts @themarpe .

themarpe:
@zrezke 
Lets keep it as is for the time being, as launcher is mainly used on windows.

Do you mind checking if launcher entry point is installed under install_depthai script? If so we can revisit for that after the fact

zrezke:
Sure!
Looks like it's present when installing depthai via: `bash -c "$(curl -fL https://docs.luxonis.com/install_depthai.sh)"` on macOS:

```sh
(x86) bash-3.2$ pwd
/Users/filipjeretina/Luxonis1/depthai
(x86) bash-3.2$ ls -la entrypoint/depthai_launcher 
-rwxr-xr-x  1 filipjeretina  staff  2650 Jun 29 13:08 entrypoint/depthai_launcher
```

zrezke:
One thing to note is that the splash screen actually crashes on mac:
```sh
2023-06-29 13:13:50.864 python[48022:399898] *** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'NSWindow drag regions should only be invalidated on the Main Thread!'
*** First throw call stack:
(
        0   CoreFoundation                      0x00007ff802f867fb __exceptionPreprocess + 242
        1   libobjc.A.dylib                     0x00007ff802ce7e58 objc_exception_throw + 48
        2   CoreFoundation                      0x00007ff802faf226 -[NSException raise] + 9
        3   AppKit                              0x00007ff80588f634 -[NSWindow(NSWindow_Theme) _postWindowNeedsToResetDragMarginsUnlessPostingDisabled] + 321
        4   AppKit                              0x00007ff80587b5e4 -[NSWindow _initContent:styleMask:backing:defer:contentView:] + 1288
        5   AppKit                              0x00007ff80587b0d6 -[NSWindow initWithContentRect:styleMask:backing:defer:] + 42
        6   AppKit                              0x00007ff805c4caeb -[_NSOrderOutAnimationProxyWindow initWithSnapshotOfWindow:] + 611
        7   AppKit                              0x00007ff805ac4c62 -[_NSWindowTransformAnimation initWithWindow:type:interruptingAnimation:] + 690
        8   AppKit                              0x00007ff805ac499b +[_NSWindowTransformAnimation windowTransformAnimationWithWindow:type:interruptingAnimation:] + 49
        9   AppKit                              0x00007ff80599e31e -[NSWindow _doOrderWindow:relativeTo:findKey:forCounter:force:isModal:] + 595
        10  AppKit                              0x00007ff80599e069 -[NSWindow orderWindow:relativeTo:] + 152
        11  libqcocoa.dylib                     0x000000011868f24a _ZN12QCocoaWindow10setVisibleEb + 1050
        12  QtGui                               0x0000000115bcd758 _ZN14QWindowPrivate10setVisibleEb + 1112
        13  QtWidgets                           0x000000011780c665 _ZN14QWidgetPrivate8hide_sysEv + 229
        14  QtWidgets                           0x000000011780e3f7 _ZN14QWidgetPrivate11hide_helperEv + 167
        15  QtWidgets                           0x000000011780e8d5 _ZN14QWidgetPrivate10setVisibleEb + 373
        16  QtWidgets.abi3.so                   0x0000000116ba47bf _ZN16sipQSplashScreen10setVisibleEb + 111
        17  QtWidgets.abi3.so                   0x0000000116d90fcb _ZL17meth_QWidget_hideP7_objectS0_ + 75
        18  python3.11                          0x000000010411ef1a cfunction_call + 90
        19  python3.11                          0x0000000104211d09 _PyEval_EvalFrameDefault + 232441
        20  python3.11                          0x00000001040bae22 _PyFunction_Vectorcall + 514
        21  python3.11                          0x00000001042187af _PyEval_EvalFrameDefault + 259743
        22  python3.11                          0x00000001040bae22 _PyFunction_Vectorcall + 514
        23  python3.11                          0x00000001040c0ef1 method_vectorcall + 449
        24  python3.11                          0x000000010430c59b thread_run + 235
        25  python3.11                          0x0000000104297374 pythread_wrapper + 36
        26  libsystem_pthread.dylib             0x00007ff802e454e1 _pthread_start + 125
        27  libsystem_pthread.dylib             0x00007ff802e40f6b thread_start + 15
)
libc++abi: terminating with uncaught exception of type NSException
```

zrezke:
linux
win
@themarpe this is ready - works on all platforms.
One UX issue is that the splash screen exits when the depthai-viewer is installing dependencies, so that might be a bit confusing for users. It should be handled on the viewer side tho. Since the viewer doesn't exit if the dependencies are not installed, one would have to parse the output to get a sense of "are we still installing dependencies?" - which is sketchy at best. I think just starting up the viewer window and then having a pop up like installing dependencies would be better. + It would be a better UX for people who pip installed it aswell...

themarpe:
> I think just starting up the viewer window and then having a pop up like installing dependencies would be better.

I agree, if Viewer is installing those dependencies, lets have that handle the Viewer specific splash screen

Erol444:
TODO: also update [Getting started with DepthAI](https://docs.luxonis.com/en/latest/pages/tutorials/first_steps/#first-steps-with-depthai) to use the Viewer by default.

zrezke:
### The dependency installer landed in depthai-viewer 0.1.4
Screenshot from 2023-07-31 16-32-43

This is fine to merge now.

themarpe (approved):
Thanks!

PR #1077, opened 2023-07-06
Support for non-standard devices

Erol444:
A few fixes:
- When creating `color` camera, it will take first color sensor, which fixes many issues for eg. OAK-D-SR
- Auto-downscale stereo frames if they're above 1280pixels in width; fixes issues with OAK-D-LR
- Select stereoLeft camera when creating `left` camera, and stereoRight when creating `right` camera. This could fix some issues with custom devices, or FFC setups, where left/right aren't cam_b/cam_c ports
- In CameraComponent, added option for `resolution='max'`, whcih will set max supported camera resolution

TODO:
- When PR: https://github.com/luxonis/depthai-python/pull/862 gets to mainline, SDK can use `Camera` node instead of `ColorCamera`/`MonoCamera` nodes.

### DIFF
--- a/depthai_sdk/src/depthai_sdk/components/camera_component.py
+++ b/depthai_sdk/src/depthai_sdk/components/camera_component.py
@@ -17,7 +17,7 @@ class CameraComponent(Component):
     def __init__(self,
                  device: dai.Device,
                  pipeline: dai.Pipeline,
-                 source: Union[str, dai.CameraBoardSocket],
+                 source: dai.CameraBoardSocket,
                  resolution: Optional[Union[
                      str, dai.ColorCameraProperties.SensorResolution, dai.MonoCameraProperties.SensorResolution
                  ]] = None,
--- a/depthai_sdk/src/depthai_sdk/components/camera_component.py
+++ b/depthai_sdk/src/depthai_sdk/components/camera_component.py
@@ -23,6 +23,7 @@ class CameraComponent(Component):
                  ]] = None,
                  fps: Optional[float] = None,
                  encode: Union[None, str, bool, dai.VideoEncoderProperties.Profile] = None,
+                 sensor_type: Optional[dai.CameraSensorType] = None,
                  rotation: Optional[int] = None,
                  replay: Optional[Replay] = None,
                  name: Optional[str] = None,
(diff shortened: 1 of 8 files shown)


OpeningNew (approved):
LGTM

depthai_sdk/src/depthai_sdk/components/camera_component.py
```diff
@@ -17,7 +17,7 @@ class CameraComponent(Component):
     def __init__(self,
                  device: dai.Device,
                  pipeline: dai.Pipeline,
-                 source: Union[str, dai.CameraBoardSocket],
+                 source: dai.CameraBoardSocket,
```
OpeningNew, line 20: I'm not sure why `source` can't be string. What if user specify "cam_b,c"?

depthai_sdk/src/depthai_sdk/oak_camera.py
```diff
@@ -131,14 +131,21 @@ def camera(self,
             encode (bool/str/Profile): Whether we want to enable video encoding (accessible via cameraComponent.out_encoded). If True, it will use MJPEG
             name (str): Name used to identify the X-out stream. This name will also be associated with the frame in the callback function.
         """
-        socket = source
         if isinstance(source, str):
-            socket = parse_camera_socket(source.split(",")[0])
+            socket_name = source.split(",")[0].lower()
```
OpeningNew, line 135: And here you replace source, so it can be only `dai.CameraBoardSocket` type. But how user can specify socket with type, like `cam_b,c`?

OpeningNew (changes requested):
What about `cam_b,c` as `source`?

OpeningNew:
I think we should split `source` (`CameraComponent`) into two args, one will represent socket, and the second type (color or mono). In the newest version of depthai they also deprecated usage of `dai.CameraBoardSocket.COLOR`, `.RGB`, `.LEFT`, etc. so it makes sense to me. This way user can explicitly say "I want color type, socket doesn't matter". Thoughts?

depthai_sdk/src/depthai_sdk/components/camera_component.py
Erol444: It gets resolved in oak_camera.py already - `c` isn't really needed anymore, as it gets resolved afterward whether it's color/mono

OpeningNew: Yep, but now user can't provide type for sensor that supports both color and mono. I guess splitting `source` into two args will solve that problem, so it works.

OpeningNew (approved):

Erol444:
@OpeningNew Added the option to specify mono/color back

OpeningNew (approved):

Jalen-Xu:
Hi,

I am running oak-d-sr camera with this latest depthai_demo sdk but I am not able to see the depth image successfully as shown below. I can only see the color image and nnInput image.

image

Erol444:
Hi @Jalen-Xu ,
the DepthAI demo doesn't have support for OAK-D-SR, and we are migrating to [depthai-viewer](https://github.com/luxonis/depthai-viewer) - I'd suggest using that one instead.

PR #1094, opened 2023-08-04
Changed camera model to pinhole (no lens distortion) for non-wide cameras

CenekAlbl:
Should be more stable for small number of calibration images.


saching13:
if `reprojection_error_threshold` is high too high then doesn't it affect the users who try to use it with pnp ?

CenekAlbl:
I needed to set higher reprojection error treshold to accomodate for lower number of camera parameters. Not sure about the value honestly - these numbers worked for the 3 cameras I had factory images for, but might still introduce failed calibrations on others. I think we should either validate with more cameras or incease it even further. What do you mean exactly by affect pnp?

PR #1104, opened 2023-08-18
Calib arguments tmp

MaticTonin:
PR created to improve calibration:

1. Improved features:

- **default board:** User can now specify just the board name to set the squareX, squareY and markerSizeCm

- **board:** user can now just connect the camera and it will automatically select the board_config file from depthai-boards. 

- **rgbLensPosition:** user can now specify camera lens position as a dictionary of multiple cameras.

- **disableRGB:** user can now specify list of cameras, which will be disabled.

-**disablePolynomsdisplay:** user can specify if they want polynoms displayed or not.

2. Features that returned:

- **invertVertical** invertVertical of frames

- **invertHorizontal** invertHorizontal of frames

- **antibanding** antibanding of lights

- **numPictures** number of pictures taken in calibration

3. Removed features:

- **numMarkers** number of markers that must be detected

- **swapLR** swaping left and right cameras

4. Also added:

- display of coverageImage while capturing images for all enabled cameras.

- stopping the capturing process with key "s"

- display error in case cameras are not in Sync

Features tested on OAK-D-PRO, OAK-D-1, OAK-D-LITE, OAK-D-PRO-W


calibrate.py
```diff
@@ -342,11 +363,46 @@ def __init__(self):
...
+            if "FF" in detection:
+                detection.remove("FF")
+            if "9782" in detection:
+                detection.remove("9782")
+            self.board_name = '-'.join(detection).upper()
+            board_path = Path(self.board_name)
+            if self.traceLevel == 1:
+                print(f"Board path specified as {board_path}")
+            if not board_path.exists():
+                board_path = (Path(__file__).parent / 'resources/depthai_boards/boards' / self.board_name.upper()).with_suffix('.json').resolve()
```
themarpe, line 390: This approach was discussed somewhere else as well. Cross apply

themarpe (changes requested):

calibrate.py
MaticTonin: Because some devices does not have name (when using the `device.getDeviceName` you get empty string) i will use try method as shown below:

```python
            try: 
                detection = self.device.getDeviceName()
                print(f"Device name: {detection}")
                detection = detection.split("-")
            except:
                cameraProperties = self.device.getConnectedCameraFeatures()
                calibData = self.device.readCalibration()
                eeprom = calibData.getEepromData()
                eeprom.productName = eeprom.productName.replace(" ", "-").upper()
                eeprom.boardName = eeprom.boardName.replace(" ", "-").upper()
                print(f"Product name: {eeprom.productName}, board name {eeprom.boardName}")
                if eeprom.productName.split("-")[0] == "OAK":
                    detection = eeprom.productName.split("-")
                elif eeprom.boardName.split("-")[0] == "OAK":
                    detection = eeprom.boardName.split("-")```
                    
Same argument. For some cameras hasAutofocus is not well defined in EEPROM or `dai.deviceInfo().`

themarpe: Mark down to be replaced with a getProductName & getDeviceName combo in v2.23 version of DepthAI
(aka if getProductName doesn't match anything, use getDeviceName)

Note, getDeviceName at the moment does what is written in the `try` case. Do you have a device where that is not the case?

MaticTonin: Ok, will do.

Yes, some old devices does not have DeviceName, but has productName or boardName. Thats why this is seperated.

PR #1117, opened 2023-09-11
Add set_exposure_iso()

Hutch07:
camera_control.py changed to add a method to set camera exposure. set_exposure_iso().  Arguments (exposure, gain).  Checks if values are outside the limits and sets values to integers.  Similar to exposure_time_down() but with values instead of steps.


depthai_sdk/src/depthai_sdk/camera_control.py
```diff
@@ -0,0 +1,378 @@
+import logging
```
Erol444, line 1: Hi @Hutch07 ,
It seems like another file (camera_control.py) was created by mistake. Could you also target `develop` branch instead of `main`?

Erol444 (changes requested):

Hutch07:
Thanks - I will

PR #1121, opened 2023-09-26
Adding debug

MaticTonin:
Adding debug mode, where camera does not need to be connected, to be calibrated. Works only for process mode.


saching13 (approved):
LGTM

PR #1130, opened 2023-10-22
Refactor encoder-related code into EncoderComponent

geektrove:
Allow users to create multiple encoders for each `CameraComponent` and `StereoComponent`

- Add `EncoderComponent`
- Ensure encoder compatible size when creating `EncoderComponent` 
- Remove encoders from `CameraComponent` and `StereoComponent`
- Remove encoded output from `NNComponent`

Tested with multiple `CameraComponent`, `StereoComponent`, recording and replaying.


Erol444:
@OpeningNew also update all examples that use videoEnc funcitonality.

PR #1151, opened 2024-02-05
Updated lib version

Erol444:
Also updated depthai-demo a bit so it doesn't throw warnings.
I don't know why windows (choco) tests fail, someone would need to look into that.


requirements.txt
```diff
@@ -6,6 +6,6 @@ depthai-sdk==1.9.4
 --extra-index-url https://artifacts.luxonis.com/artifactory/luxonis-depthai-data-local/wheels/
 pyqt5>5,<5.15.6 ; platform_machine != "armv6l" and platform_machine != "armv7l" and platform_machine != "aarch64" and platform_machine != "arm64"
 --extra-index-url https://artifacts.luxonis.com/artifactory/luxonis-python-snapshot-local/
-depthai==2.21.2.0
+depthai==2.24.0.0
```
jakaskerl, line 9: This is fine, but won't work with python3.12 until the next official depthai release. Also depthai_sdk requirement will try to force install latest depthai (which fails on 312) first, so manually installing depthai using artifactory won't fix the issue.

jakaskerl (approved):
LGTM :)

PR #1154, opened 2024-02-09
Allow use of prior cached model

dlaflotte:
Moving the requests.get(url) below the check for the cache.  If your vision system is offline, and you have a cached model, the call to requests prior to checking the cache throws an exception.  Now if you have a cached model it will use that first and only if you do not have a model will it switch to checking api.roboflow.com


Erol444 (approved):

Erol444:
That seems like an useful fix, thanks for the PR!

PR #1182, opened 2024-06-25
SDK 1.15

Erol444:
Added ToF configuration function, support ToF depth from pointcloud, pin numpy below 2.0, depthai-sdk to 1.15

### DIFF
--- a/depthai_sdk/src/depthai_sdk/components/pointcloud_component.py
+++ b/depthai_sdk/src/depthai_sdk/components/pointcloud_component.py
@@ -37,11 +37,11 @@ class PointcloudComponent(Component):
 
         # Depth aspect
         if depth_input is None:
-            stereo = StereoComponent(device, pipeline, replay=replay, args=args)
-            stereo.config_stereo(lr_check=True, subpixel=True, subpixel_bits=3, confidence=230)
-            stereo.node.initialConfig.setNumInvalidateEdgePixels(20)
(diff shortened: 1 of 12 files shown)


depthai_sdk/src/depthai_sdk/components/pointcloud_component.py
```diff
@@ -58,15 +58,18 @@ def __init__(self,
...
-            stereo = stereo.node
-
-        if isinstance(stereo, dai.node.StereoDepth):
-            self.stereo_depth_node = stereo
-            self.depth = stereo.depth
-        elif isinstance(stereo, dai.Node.Output):
-            self.stereo_depth_node = stereo.getParent()
-            self.depth = stereo
+        if isinstance(depth_input, StereoComponent):
+            depth_input = stereo.node
```
jakaskerl, line 62: depth_input = depth_input.node
self.stereo_depth_node = depthai_input.getParent()
....
...

jakaskerl (commented):
stock pointcloud example is not working, because of the changed parameter name.

jakaskerl:
I see we have 2 requirements files, one in base depthai dir and one in sdk. The first one needs `numpy` version lock. Would it make sense to put them into a single file and have `install_requirements.py` as we have in `depthai-python`?

jakaskerl (approved):
lgtm

PR #1228, opened 2025-02-20
SDK offline NN

jakaskerl:
Made it so SDK (=> viewer) doesn't crash if no internet access is available.


Erol444 (approved):

zrezke:
Just wondering what would happen now if you select a NN model to run in the viewer, when you're offline @jakaskerl ?

jakaskerl:
@zrezke Works as intended because the models are cached on first install.

zrezke:
Good! thanks @jakaskerl !

zrezke (approved):

